PR #9, opened 2020-07-27
Fix minimization

mwillsey:
Turns out pairing equalities together just isn't the right abstraction.


mwillsey:
Correct, there is now no such thing as a double sided rewrite. You add both sides separately. The minimizer can remove one or both as it sees fit. Now the minimizer only adds the lhs of a rule and tests for the rhs. This is better than before, but still not foolproof, since the RHS of the pattern could still possibly be inserted by some other pattern.

PR #25, opened 2021-09-21
Modulo alpha renaming

bksaiki:
This PR implements enumeration modulo alpha (EMA), modifies `choose_eqs()`, and removes some old code.


src/bin/rational.rs
```diff
@@ -175,27 +175,139 @@ impl SynthLanguage for Math {
...
             }
-            to_add.push(Math::Abs(i));
-            to_add.push(Math::Neg(i));
         }
 
         log::info!("Made a layer of {} enodes", to_add.len());
         to_add
     }
 
+    fn valid_constants(
```
chandrakananandi, line 217: I haven't looked at the entire PR yet, but this seems tricky to me ---  traversing the egraph recursively can be very expensive so we should discuss if we really need this or there is a way to accomplish similar effects differently.

I will comment more on the PR later. Looks fantastic so far!!

src/lib.rs
```diff
@@ -182,6 +205,41 @@ pub trait SynthLanguage: egg::Language + Send + Sync + 'static {
...
+        _id: &Id,
+        _seen: &mut HashSet<Id>
+    ) -> bool {
+        true
+    }
+
+    /// Given an eclass `id`, returns true if any id in `ids` is
+    /// encountered when recursing down `depth` times.
+    /// Used for aggressive constant filtering
+    fn contains_ids_with_depth(
```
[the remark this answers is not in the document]
chandrakananandi: Also, I am not following the intuition here. Maybe should discuss in next meeting?

src/lib.rs
```diff
@@ -492,35 +546,76 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
+            // deduplicate and filter bad constants
+            let mut cp = self.egraph.clone();
+            let mut max_id = cp.number_of_classes();
+            layer.retain(|node| {
+                let seen = &mut HashSet::<Id>::default();
+                if iter > self.params.ema_above_iter {
+                    let rec = node.to_recexpr(|id| self.egraph[id].data.simplest.as_ref());
+                    let rec2 = L::emt_generalize(&rec);
+                    let id = cp.add_expr(&rec2);
+                    if usize::from(id) < max_id {
```
chandrakananandi, line 579: umm, what does this check do?

src/lib.rs
```diff
@@ -993,133 +1162,53 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
-            }
-            let n_rules = usize::min(self.params.rules_to_take, new_eqs.len());
-            if step < 10 && n_rules > 200 {
-                break;
+        let mut step_idx = 0;
+
+        // Idea here is to remain at a high step level as long as possible
+        // Move down in step size if
+        //  (a) the number of eqs is smaller than the step size
+        //  (b) the number of bad eqs is smaller than the step size
```
chandrakananandi, line 1177: why b?

src/lib.rs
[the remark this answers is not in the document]
bksaiki: Also killed this feature. Kind of nonsensical and too aggressive.

src/lib.rs
bksaiki: I had a reason before, but I'm blanking on why that's the case. We can discuss this change.

Cargo.toml
```diff
@@ -21,8 +21,8 @@ lazy_static = "1.4.0"
 itertools = "0.9.0"
 # fixed = "1.9.0"
 clap = "3.0.0-beta.2"
-z3 = {version="0.10.0", features = ["static-link-z3"]}
-# z3 = {version="0.10.0"}
+#  z3 = {version="0.10.0", features = ["static-link-z3"]}
+z3 = {version="0.10.0"}
```
bksaiki, line 25: Ooof, no. Reverting this. Honestly, this is the worst part of the build process. The build time with the static-link is super long, but its unreasonable to assume people have `libz3` installed on their system.

src/bin/rational.rs
```diff
@@ -175,27 +175,139 @@ impl SynthLanguage for Math {
...
+                    synth.egraph.number_of_classes());
+
         for i in synth.ids() {
             for j in synth.ids() {
-                if synth.egraph[i].data.exact && synth.egraph[j].data.exact {
-                    continue;
+                // must have at least one non-constant term
+                // and at least one term from previous iteration
+                if synth.egraph[i].data.gen + 1 == _iter || synth.egraph[j].data.gen + 1 == _iter {
+                    if _iter > synth.params.no_constants_above_iter {
```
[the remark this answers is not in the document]
bksaiki: Well, since the change from size to depth isn't in `main`, I suppose this can stay for now. It'll just be overwritten.

src/lib.rs
bksaiki: Removes a newly enumerated node if it is redundant. Yes the egraph will automatically do this, but deduplicating nodes before cvec-matching and minimization is helpful since minimization takes a long time.

Cargo.toml
chandrakananandi: ugh yeah sorry about this. It's annoying for sure.

src/bin/rational.rs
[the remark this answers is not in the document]
chandrakananandi: ok yeah sounds great.

src/lib.rs
chandrakananandi: huh that's very interesting! Cool!

src/lib.rs
chandrakananandi: Ok yeah just explain to me sometime this week. I will resolve this for now.

PR #30, opened 2022-01-25
Add CI with GitHub Actions

bksaiki:
This PR adds CI to Ruler. The workflow runs Rust unit tests, synthesizes rules for a few domains, and compares their proving power against previously generated rule sets. In addition, the build process will no longer statically link to Z3, so users need to install `libz3` on their system (users can manually edit the `Cargo.toml` to revert this behavior). The README has been simplified and updated to reflect these changes while the original OOPSLA documentation now in the OOPSLA21.md file.


.github/workflows/tests.yml
```diff
@@ -0,0 +1,72 @@
...
+
+on: [push]
+
+env:
+  RUST_BACKTRACE: full
+
+jobs:
+  unit:
+    name: "Unit Tests"
+    runs-on: ubuntu-latest
```
chandrakananandi, line 11: is this what warfa has?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,72 @@
...
+
+      - name: "Build"
+        run: cargo build --release
+
+      - name: "Unit tests"
+        run: cargo test --release
+
+      - name: "Synthesize rational rules"
+        run: |
+          cargo rational --iters 2 --variables 3 --num-fuzz 50 --do-final-run --outfile rat.json
```
chandrakananandi, line 42: i think it's worth saving the result jsons in a designated directory.

chandrakananandi: I know there is `tests/` but that's the expected ones right? What if we make a `tests/expected` and `tests/actual`? Thoughts?

src/bin/real-rat.rs
```diff
@@ -118,7 +166,7 @@ define_language! {
         "-R" = RSub([Id; 2]),
         "*R" = RMul([Id; 2]),
         "/R" = RDiv([Id; 2]),
-        Var(egg::Symbol),
+        Var(Variable),
```
chandrakananandi, line 169: why this change?

src/derive.rs
```diff
@@ -9,45 +9,68 @@ pub fn parse<L: SynthLanguage>(filename: &String) -> Vec<Pair<L>> {
...
-            "not_derivable": pairs_to_eqs(&rev_not_derivable),
-        },
-    });
+    let (pairs1, pairs2) = if params.new_eqs {
+        (parse_new_eqs::<L>(&params.in1), parse_new_eqs::<L>(&params.in2))
+    } else {
+        (parse::<L>(&params.in1), parse::<L>(&params.in2))
+    };
+
+    if params.ci {
```
chandrakananandi, line 40: yup this seems good!

chandrakananandi (commented):
left some questions and comments.

.github/workflows/tests.yml
chandrakananandi: or maybe we can call it `out/` for the new ones and `tests/` for the pregenerated ones..

scripts/generate-tests.sh
```diff
@@ -0,0 +1,28 @@
+#!/usr/bin/env bash
```
chandrakananandi, line 1: where is this script run? Just once to generate the "original" results I suppose?

.github/workflows/tests.yml
bksaiki: This is just CI, not nightlies. We would need a considerably more complicated system for that.

.github/workflows/tests.yml
bksaiki: Makes sense to me. I'll probably go with `out/` and `tests/`.

scripts/generate-tests.sh
bksaiki: Yep. It should only be rerun, if we want to change our expected (baseline) rulesets.

.github/workflows/tests.yml
chandrakananandi: oh right sorry I was still thinking of nightlies in my head ever since we discussed it in the egraph channel.

src/bin/real-rat.rs
bksaiki: Parsing issue. It was actually causing derive to crash.

chandrakananandi: weird, ok!

chandrakananandi (approved):
Thanks Brett!!

PR #33, opened 2022-02-08
Rollback egg version

bksaiki:
This PR reverts the egg version to a pre-proof commit (8d68c8768cc18c2eae97f970e2b4162324cee787). The introduction of delayed unions and proofs in egg 0.7.x has led to mysterious problems within Ruler that have been difficult to diagnose. Until a solution can be found, using an older version of egg will have to suffice.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -33,17 +33,17 @@ jobs:
 
       - name: "Synthesize rational rules"
         run: |
-          cargo rational --iters 2 --variables 3 --num-fuzz 50 --do-final-run --outfile out/rat.json
+          cargo rational --iters 2 --variables 3 --use-smt --do-final-run --outfile out/rat.json
           cargo run --release --bin rational derive --ci out/rat.json tests/rat.json out/rat-derive.json
 
       - name: "Synthesize rational rules (EMA)"
         run: |
-          cargo rational --iters 2 --variables 3 --ema-above-iter 1 --num-fuzz 50 --do-final-run --outfile out/rat-ema.json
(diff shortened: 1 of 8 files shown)


bksaiki:
Synthesizer tests have been modified due to potential issues with egg 0.7. Things to note:
- `rational` rules have decreased in proving power (cause: switched to `--use-smt` rather than `--num-fuzz`)
- `bv`<n> rules are larger by 3 or 4 rules than test rules (when did this sneak in?)
- `real` (rule lifting) now fixed (test rules had unsound rules),
- `real` lost ability to prove `(*R 0R ?a) => 0R` and `(/R 0R ?a) => 0R` (but this is fixed when the rule lifting algorithm is updated to match `cleanup` or `rule_lifting` branch (test rules should be updated later)

bksaiki:
Update: increase in number of bv rules seems to be from reverting the egg version

bksaiki:
bv32 produces one less rule than the `oopsla21-aec` branch with the following derivability result:
```
Using bv32.json to derive bv32-oopsla.json
44 rules are derivable, 2 are not.

> (* ?a (-- ?b ?b)) => (-- ?a ?a)
> (<< (-- ?a ?a) ?b) => (-- ?b ?b)

Using bv32-oopsla.json to derive bv32.json
45 rules are derivable, 2 are not.

> (<< (-- ?a ?a) ?b) => (-- ?a ?a)
> (* ?a (-- ?b ?b)) => (-- ?b ?b)
```
These look like identical rules just the variable naming is altered. Might be worth looking into, but for now I think this branch is ready for merging.

chandrakananandi (approved):
yes sounds good to me! Thanks for checking against oopsla-aec.

PR #34, opened 2022-02-08
Add unit tests for bool

ajpal:
Unit tests ensuring we find the exact right rulesets for bool for one and two iters


chandrakananandi:
This seems to have the egg version issue. I suggest that Brett's PR to fix the egg version gets merged first, then you can merge `main` to your branch to fix the egg issue. That should hopefully work!

bksaiki (approved):
Minor suggestion: the error message for an unknown rule is not very helpful:
```
Learned 12 rules in 0.003377525 using 0 old rules.
thread 'test::iter1_rules' panicked at 'assertion failed: expected.contains(&rule.to_string().as_str())', src/bin/bool.rs:209:30
stack backtrace:
```
It doesn't say what rule is not found in the set. Otherwise, looks good. Looking forward to more tests.

PR #35, opened 2022-02-09
Add rustfmt and clippy to CI

ajpal:
1st commit adds to the Github Actions
2nd commit is just the result of runnning `cargo fmt --all` so probably doesn't need careful review
3rd commit is manual changes to get clippy passing- code review would be helpful here.

I manually checked that things still work for bool with one and two iters, but let me know if there are other manual checks that would be helpful to make sure this didn't inadvertently break anything


bksaiki (approved):
Looks good to me!

PR #37, opened 2022-02-11
Remove the valid constant checks and split the CI jobs

chandrakananandi:
1. removes the `valid_constants` filtering. Seems to not affect results.
2. split the jobs into unit tests and regression tests in CI.

I am not sure what is leading to some flakiness in the CI. I had to rerun the jobs to get them to pass. Any idea?


bksaiki (approved):
Much cleaner. I like it! If we every want this feature in the future, we should think of a saner way to do it.

PR #39, opened 2022-02-12
Cleanup domains and synth loop

bksaiki:
This PR is an amalgamation of structural changes and refactoring including:
 - break up synthesis loop into multiple functions
 - make rule lifting loop structurally similar to cvec-based synth loop
 - eliminate unnecessary code (everywhere)
 - fix rule lifting steps to match high-level algorithm
 - various fixes
 - fix div-by-zero "validation" in real, complex domains


src/bin/real-rat.rs
```diff
@@ -268,6 +252,19 @@ impl SynthLanguage for Math {
         )
     }
 
+    fn is_extractable(&self) -> bool {
```
chandrakananandi, line 255: same as other lifted domains. I think this is actually a better name than is_in_domain FWIW

src/lib.rs
```diff
@@ -51,12 +50,29 @@ pub enum ConstantFoldMethod {
...
 
+/// Validation result
+#[derive(Debug, Clone)]
+pub enum ValidationResult {
+    Valid,
+    Invalid,
+    Unknown,
+}
+
+impl From<bool> for ValidationResult {
```
chandrakananandi, line 61: I like this!

src/lib.rs
```diff
@@ -819,187 +835,197 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
-                    self.egraph.total_size(),
-                    self.egraph.number_of_classes(),
-                );
+        // run lifting rules
+        log::info!("running lifting rules");
+        let runner = self
+            .mk_cvec_less_runner(self.egraph.clone())
+            .with_iter_limit(self.params.eqsat_iter_limit + 1); // increase iter limit by 1 to ensure saturation
+        let rewrites: Vec<&Rewrite<L, SynthAnalysis>> = self.lifting_rewrites.iter().collect();
+        let (new_egraph, found_unions) = self.run_rewrites_with_unions(rewrites, runner);
```
chandrakananandi, line 851: new_egraph is the dirty egraph here?

chandrakananandi: why does `run_rewrites_with_unions` return the egraph at all?

src/lib.rs
```diff
@@ -819,187 +835,197 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
+                {
+                    let extract = Extractor::new(&self.egraph, ExtractableAstSize);
+                    let (_, e1) = extract.find_best(win[0]);
+                    let (_, e2) = extract.find_best(win[1]);
+                    if let Some(mut eq) = Equality::new(&e1, &e2) {
+                        if e1 != e2 &&
+                            e1.as_ref().iter().all(|x| x.is_extractable()) &&        // extractable and valid
+                            e2.as_ref().iter().all(|x| x.is_extractable())
+                        {
+                            if let ValidationResult::Valid = L::validate(self, &eq.lhs, &eq.rhs) {
```
chandrakananandi, line 870: why this and not `is_valid_rewrite`? Why do we need is_valid_rewrite again?

src/lib.rs
```diff
@@ -819,187 +835,197 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
 
-                // run new eqs
-                log::info!("running eqsat with {} new rules", self.new_eqs.len());
-                let mut runner = self.mk_cvec_less_runner(self.egraph.clone());
-                let rewrites: Vec<&Rewrite<L, SynthAnalysis>> =
-                    self.new_eqs.values().flat_map(|eq| &eq.rewrites).collect();
-                runner = runner.run(rewrites);
-                self.egraph = runner.egraph;
-                self.egraph.rebuild();
+        self.egraph = new_egraph;
```
chandrakananandi, line 883: CONFUSION!! why do we do this again?

src/lib.rs
```diff
@@ -819,187 +835,197 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
-                let mut new_eqs: EqualityMap<L> = EqualityMap::default();
-                for ids in found_unions.values() {
-                    for win in ids.windows(2) {
-                        if win[0] != win[1]
-                            && self.egraph[win[0]].data.is_extractable
-                            && self.egraph[win[1]].data.is_extractable
+        self.egraph.rebuild();
+
+        // run forbidden rules
+        log::info!("running forbidden rules");
```
chandrakananandi, line 901: this is where we learn rules in the allowed domain, right?

src/lib.rs
```diff
@@ -819,187 +835,197 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
-                let (_, e) = extract.find_best(id);
-                log::info!(
-                    "{} [{}]: {:?} {:?}",
-                    id,
-                    self.egraph[id].data.is_extractable,
-                    e.pretty(100),
-                    self.egraph[id].nodes
-                );
-            }
+        // final run_rewrites
```
chandrakananandi, line 1020: does this still make a difference?

chandrakananandi (changes requested):
I have some questions and suggestions. My main confusion is about where we run rules on a cloned egraph and where we do it directly on the pristine one. It would help if we can find all places where we run rewrites and write down our strategy for each one and a justification for it. 

Thanks for all the cleanup Brett! Looks great in general!

.github/workflows/tests.yml
```diff
@@ -115,6 +115,6 @@ jobs:
           cargo real-rat --iters 2 --variables 3 --no-constants-above-iter 1 --prior-rules out/rat.json --outfile out/real.json
           cargo run --release --bin real-rat derive --ci --new-eqs out/real.json tests/real.json out/real-derive.json
 
-      # Need to filter out rational rules from real rules
+      # Need to filter out rational rules from real.json
```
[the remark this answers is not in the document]
bksaiki: The rules it would pull in also includes rational ops which complex can't parse.

.gitignore
```diff
@@ -2,6 +2,7 @@
 /target
 /out
 /*.json
+/*.sh
```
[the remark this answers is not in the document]
bksaiki: Yep. I believe that's the right syntax for the top level. This could easily be changed in the future.

src/bin/bigint.rs
```diff
@@ -160,11 +160,11 @@ impl SynthLanguage for Math {
         to_add
     }
 
-    fn is_valid(
+    fn validate(
```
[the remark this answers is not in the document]
bksaiki: Actually, that enum is just an artifact from some design choice I removed. Honestly, an `Option<bool>` might be a better idea.

src/bin/bv-bool.rs
```diff
@@ -112,12 +86,13 @@ impl SynthLanguage for Math {
     fn is_in_domain(&self) -> bool {
```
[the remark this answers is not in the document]
bksaiki: I really need to rethink the `is_in_domain` method and `in_domain` field of the analysis. I feel like this should be made into an optional typing function. Something like `type_id() ->  usize` that can be used for debugging to make sure eclasses of different "types" do not merge.

src/lib.rs
bksaiki: Bad naming convention. I believe that `validate()` returns whether or not a rule is valid to return to the user while `is_valid_rewrite` checks if the rule should be applied.

src/lib.rs
bksaiki: Probably not

src/lib.rs
bksaiki: Lifting => the runner egraph is a denoted version, we want to keep the extra nodes

src/lib.rs
```diff
@@ -819,187 +835,197 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
-                runner = runner.run(&self.lifting_rewrites);
-
-                // collect any interesting unions
-                log::info!("{:?} collecting unions...", runner.stop_reason.unwrap());
-                let mut found_unions = HashMap::default();
-                for id in self.ids() {
-                    let id2 = runner.egraph.find(id);
-                    found_unions.entry(id2).or_insert_with(Vec::new).push(id);
-                }
+        // run allowed rules
```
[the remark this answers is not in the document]
bksaiki: Allowed rules functions like `run_rewrites` in the cvec-based approach. It just compacts the egraph.

src/bin/bv-bool.rs
```diff
@@ -34,7 +34,7 @@ define_language! {
 
 fn is_bv_str(s: &'static str) -> impl Fn(&mut EGraph<Math, SynthAnalysis>, Id, &Subst) -> bool {
     let var = s.parse().unwrap();
-    move |egraph, _, subst| egraph[subst[var]].data.in_domain
+    move |egraph, _, subst| egraph[subst[var]].data.is_allowed
```
chandrakananandi, line 37: much better actually

src/lib.rs
```diff
@@ -842,17 +842,43 @@ impl<L: SynthLanguage> Synthesizer<L> {
         let mut new_candidate_eqs: EqualityMap<L> = EqualityMap::default();
         let run_rewrites_before = Instant::now();
 
-        // run lifting rules
+        // 1) Allowed rules
```
chandrakananandi, line 845: thanks for the comments!!

chandrakananandi (approved):

PR #40, opened 2022-02-23
Rational interval analysis

ajpal:
For some reason when I add tests they pass locally but then the build step fails in CI? Has anyone seen that before? I'll try to add tests as a separate PR to help isolate the issue

### DIFF
--- a/src/bin/rational.rs
+++ b/src/bin/rational.rs
@@ -9,7 +9,7 @@
    - implement `is_valid` for checking the validity of the rules in your domain.
 !*/
 
-use std::ops::{Add, Neg};
+use std::ops::{Add, Mul, Neg};
 
 use egg::*;
 use ruler::*;
@@ -23,6 +23,12 @@ use z3::*;
 
 /// define `Constant` for rationals.
 pub type Constant = Ratio<BigInt>;
+
+enum Sign {
+    Positive,
+    ContainsZero,
+    Negative,
+}
 
 define_language! {
     /// Define the operators for the domain.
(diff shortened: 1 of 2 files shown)


bksaiki:
Chandra and I have been having trouble with this bug. For some odd reason the rust compiler will just sporadically fail. I suggest just re-running the CI job when this happens.

src/bin/rational.rs
```diff
@@ -47,6 +49,41 @@ fn mk_constant(n: &BigInt, d: &BigInt) -> Option<Constant> {
...
+fn neg(interval: Interval<Constant>) -> Interval<Constant> {
+    (interval.1.map(|x| x.neg()), interval.0.map(|x| x.neg()))
+}
+
+fn abs(interval: Interval<Constant>) -> Interval<Constant> {
+    let (min, max) = interval;
+    let zero = Ratio::new(0.to_bigint().unwrap(), 1.to_bigint().unwrap());
+
+    let new_min = match (min.clone(), max.clone()) {
+        (None, None) => zero,
```
chandrakananandi, line 61: what does `None` encode in these intervals? I thought it was `-inf` and `+inf` depending on which side of the interval it is on. Is that the case? We should add that in a comment if so.

chandrakananandi: not here, but probably in `lib.rs` where Interval is defined.

chandrakananandi (changes requested):
I left just a few suggestions for the data structure to use for `Interval` and questions about the use of `None` but looks good to me over all!

src/bin/rational.rs
```diff
@@ -47,6 +49,41 @@ fn mk_constant(n: &BigInt, d: &BigInt) -> Option<Constant> {
...
+
+    let new_max = match (min, max) {
+        (Some(a), Some(b)) => Some(a.abs().max(b.abs())),
+        _ => None,
+    };
+    (Some(new_min), new_max)
+}
+
+fn add(a: Interval<Constant>, b: Interval<Constant>) -> Interval<Constant> {
+    let add = |(a, b): (Constant, Constant)| Some(a.add(b));
```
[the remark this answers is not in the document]
ajpal: oh, I didn't realize `+` was overloaded :)

chandrakananandi (approved):

PR #42, opened 2022-02-24
Mul + Recip + Div interval analysis

ajpal:
Ignore the first commit- that's https://github.com/uwplse/ruler/pull/41

### DIFF
--- a/src/bin/rational.rs
+++ b/src/bin/rational.rs
@@ -479,3 +479,70 @@ fn egg_to_z3<'a>(
 fn main() {
     Math::main()
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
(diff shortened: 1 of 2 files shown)


bksaiki (approved):
This looks good. I think the only todo that can be implemented is `mul` for `[-a, b] x [-c, d]`. I'm glad we have some implementation for div!

bksaiki:
You could actually finish that `mul` case by computing `[-a, 0] x [-c, d]` and `[0, b] x [-c, d]` separately and taking the interval that bounds both of those subintervals.

chandrakananandi (approved):
Looks good to me! I think @bksaiki has some suggestions for the last case of `mul` but rest makes sense! Thanks Anjali!!

PR #43, opened 2022-02-24
Better last case of mul

ajpal:
Let interval 1 be `(a, b)` and interval 2 be `(c, d)` where `a,c < 0` and `b,d > 0`
The potential endpoints of the interval are: `ac (+), ad (-), bc (-), bd (+)`
So the low end of the new interval is `min(ad, bc)` and the high end is `max(ac, bd)`
If any of the endpoints are `inf`, the new interval is always `(-inf, inf)`


chandrakananandi (approved):
Looks great, thanks for the explanation in the commit description!

chandrakananandi:
Merged since @bksaiki already discussed the math with Anjali in the previous PR.

PR #44, opened 2022-02-24
Add interval analysis option for constant folding

ajpal:
I'm not sure the best way to test this, let me know if you have suggestions.
I manually checked by printing out each fold, and this finds the exact same folds as CvecMatching for iters 1 and 2 for rationals


bksaiki:
That is tough. I'm assuming it's good if it doesn't affect our expect tests.

bksaiki (commented):
Other than the extraneous `println!`, looks good to me!

bksaiki (approved):
Sorry! I meant to choose approve.

chandrakananandi (approved):
Very glad about this! :shipit:

PR #45, opened 2022-02-24
Bool interval analysis for constant folding

ajpal:
For some reason, this folds a lot more expressions than cvec matching. I think it has something to do with `data.exact` but I don't totally understand it yet. I'm still looking into it, but thought I'd open the PR for discussion if you have any ideas.

Edit: It looks like the difference is coming from terms like `(and false ?x)`. The interval for this will be `(false, false)` so it gets constant folded if you're using IntervalAnalysis, but it's not an `exact` term, so it won't get constant folding if you're using CvecMatching. This also explains why it didn't come up for rational (none of the ops short-circuit on their args the same way). In that case, I think this interval analysis is good, and strictly better than the old cvec matching constant fold?

### DIFF
--- a/src/bin/bool.rs
+++ b/src/bin/bool.rs
@@ -49,6 +49,77 @@ impl SynthLanguage for Math {
         }
     }
 
+    fn mk_interval(&self, egraph: &EGraph<Self, SynthAnalysis>) -> Interval<Self::Constant> {
+        let mut interval = Interval::default();
+        match self {
+            Math::Lit(n) => interval = Interval::new(Some(*n), Some(*n)),
+            Math::Var(_) => interval = Interval::new(Some(false), Some(true)),
+            Math::Not(x) => {
+                if let Interval {
+                    low: Some(a),
+                    high: Some(b),
+                } = egraph[*x].data.interval.clone()
+                {
+                    interval = Interval::new(Some(!b), Some(!a))
+                }
+            }
+            Math::And([x, y]) => {
+                if let Interval {
+                    low: Some(a),
+                    high: Some(b),
+                } = egraph[*x].data.interval.clone()
+                {
+                    if let Interval {
+                        low: Some(c),
+                        high: Some(d),
+                    } = egraph[*y].data.interval.clone()
+                    {
+                        interval = Interval::new(Some(a && c), Some(b && d))
+                    }
+                }
+            }
+            Math::Or([x, y]) => {
+                if let Interval {
+                    low: Some(a),
+                    high: Some(b),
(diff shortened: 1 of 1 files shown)


bksaiki (approved):
All the logic looks good to me. It's great that this improves constant folding. Good find!

chandrakananandi:
> For some reason, this folds a lot more expressions than cvec matching. I think it has something to do with `data.exact` but I don't totally understand it yet. I'm still looking into it, but thought I'd open the PR for discussion if you have any ideas.
> 
> Edit: It looks like the difference is coming from terms like `(and false ?x)`. The interval for this will be `(false, false)` so it gets constant folded if you're using IntervalAnalysis, but it's not an `exact` term, so it won't get constant folding if you're using CvecMatching. This also explains why it didn't come up for rational (none of the ops short-circuit on their args the same way). In that case, I think this interval analysis is good, and strictly better than the old cvec matching constant fold?

@ajpal how are you comparing the two approaches?

chandrakananandi:
> > For some reason, this folds a lot more expressions than cvec matching. I think it has something to do with `data.exact` but I don't totally understand it yet. I'm still looking into it, but thought I'd open the PR for discussion if you have any ideas.
> > Edit: It looks like the difference is coming from terms like `(and false ?x)`. The interval for this will be `(false, false)` so it gets constant folded if you're using IntervalAnalysis, but it's not an `exact` term, so it won't get constant folding if you're using CvecMatching. This also explains why it didn't come up for rational (none of the ops short-circuit on their args the same way). In that case, I think this interval analysis is good, and strictly better than the old cvec matching constant fold?
> 
> @ajpal how are you comparing the two approaches?

Strong agree that this approach is better than `cvec-matching`. I don't think `cvec-matching` is a good way, it works out but should not be preferred.

chandrakananandi (approved):
Logic looks good to me! I am curious how you measured the two approaches though.

ajpal:
@chandrakananandi Nothing fancy, I just ran it with both folding methods and used `println` to count how many times it folded things. Then, since it was clear that Interval Analysis was folding way more terms, I printed out a little more information in the cases where the interval analysis folded but the cvec matching wouldn't have: 
```
if !sig.exact {
                            let extract = Extractor::new(&egraph, ExtractableAstSize);
                            println!("interval fold: {:?}", a);
                            let x = extract.find_best(id);
                            println!("{:?}", x);
                            println!("{:?}", sig.cvec);
                        }
```
That's how I figured out it was terms like `(and false ?x)` that were causing the difference

chandrakananandi:
Very cool!! IA is strictly better that the other one.

PR #46, opened 2022-02-25
Bug fix: Only update interval if it changed

ajpal:
Bug pointed out by @bksaiki - we should only be updating `to.interval` if it actually changed, and we should set `merge_a` to true when that happens.


bksaiki (approved):
This looks correct! We'll see what CI says.

PR #47, opened 2022-03-02
minor tweak to reports to avoid parsing issues when lifting > 1

chandrakananandi:
Right now if you want to run ruler for a doubly lifted domain like complex-real, you have to manually remove the rules over rationals since they are not part of the grammar for complex (real and complex ops are). With this change, that will no longer be needed.

There are a few caveats however:
- for lifted rule inference, you can't see the old ruleset used to learn the new rules
- `all_eqs` is therefore just `new_eqs` and `old_eqs = []`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1058,7 +1058,6 @@ impl<L: SynthLanguage> Synthesizer<L> {
         n_eqs.sort_by_key(|eq| eq.score());
         n_eqs.reverse();
         let num_olds = self.old_eqs.len();
-        let o_eqs: Vec<_> = self.old_eqs.clone().into_iter().map(|(_, eq)| eq).collect();
         for eq in &n_eqs {
             println!("  {:?}   {}", eq.score(), eq);
         }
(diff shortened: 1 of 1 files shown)


bksaiki:
I think this prevents us from running in the same domain multiple times. For example, imagine running the rational domain 5 times each with different parameters. You'll only be able to save rules for one additional run since they'll be thrown out after the next run.

chandrakananandi:
> I think this prevents us from running in the same domain multiple times. For example, imagine running the rational domain 5 times each with different parameters. You'll only be able to save rules for one additional run since they'll be thrown out after the next run.

@bksaiki is that true? Note that I changed the report only for `run_rule_lifting` and not for `run_cvec_synth`. Another thing to note is that in the body of `pub fn new(params: SynthParams) { ... }` when `old_eqs` is populated, it is actually taking _**all**_ the rules from the prior_rules file, not just the old rules. I may be wrong here in my reasoning so let's double or even triple check this.

PR #48, opened 2022-03-07
Enable running from a workload

ajpal:
2 main changes:
1. `enumerate_layer` now returns a `Vec<RecExpr<L>>` instead of `Vec<L>` and `add_chunk`/`run_chunk` expect `&[RecExpr<L>]` instead of `&[L]`. This is because we parse the workload terms into `RecExpr<L>` so it simplifies the code to make everything the same type.
2. If you set iters to 0 and supply a workload file, Ruler will run with the workload terms instead of standard enumeration.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -637,34 +637,25 @@ impl<L: SynthLanguage> Synthesizer<L> {
         // deduplicate
         let mut cp = self.egraph.clone();
         let mut max_id = cp.number_of_classes();
-        layer.retain(|node| {
-            if iter > self.params.ema_above_iter {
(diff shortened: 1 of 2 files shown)


bksaiki (approved):
I see what the confusion was over EMA now. This is much cleaner!

src/lib.rs
```diff
@@ -658,23 +658,28 @@ impl<L: SynthLanguage> Synthesizer<L> {
         });
 
         layer
+            .iter()
```
chandrakananandi, line 661: why do we need this on top of the deduplication code right above?

chandrakananandi: you could use a filter_map as Brett was saying.

chandrakananandi (approved):

PR #49, opened 2022-03-13
Make iters and workload conflicting args

ajpal:
Either workload or iters should be provided, not both

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -800,13 +800,17 @@ impl<L: SynthLanguage> Synthesizer<L> {
         let mut poison_rules: HashSet<Equality<L>> = HashSet::default();
         let t = Instant::now();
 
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1044,8 +1048,9 @@ impl<L: SynthLanguage> Synthesizer<L> {
         runner.egraph.rebuild();
         self.egraph = runner.egraph;
 
-        assert!(self.params.iters > 0);
-        for iter in 1..=self.params.iters {
+        let iters = self.params.iters.expect("iters required for rule lifting");
```
chandrakananandi, line 1051: this is more a question for @bksaiki: didn't you say you were using custom workloads for trig? How does that work with iters?

chandrakananandi (approved):

PR #57, opened 2022-04-19
Persistent cvec implementation.

vee-effekt:
This PR addresses #51 over the rational domain.

Key changes:
- When a bad rule is disproven via fuzzing, the assignment that disproved that rule--a counterexample--is written to a file.
- A user-specified number of counterexamples can be deserialized from the counterexamples file and added to the cvec upon initialization.
- Serialization and deserialization of counterexamples is toggleable from the command line.


vee-effekt:
Noticed a bug, oops.

vee-effekt:
Fixed! Also: observation, when loading ONLY counterexamples into the cvec, we find more rules in approximately the same amount of time. Weird!!!!!

amyjzhu:
Thanks for this Thia!! My thought would be that before this goes to main, we might want to generalize some of it (e.g. move the definition of Counterexample, etc. into lib, as well as the loading/writing of counterexample files)

vee-effekt:
> Thanks for this Thia!! I think before this goes to main, we might want to generalize some of it (e.g. move the definition of Counterexample, etc. into lib, as well as the loading/writing of counterexample files). Would like to hear others' thoughts on this too

Yes, so would I! I can definitely move the new structs into lib. Writing counterexamples is trickier since both init_synth and validate have domain-specific implementations. validate for example gets called in lib and returns a ValidationResult rather than a vector representing the output of the fuzzer, so I'd need to get that output somehow. The obvious way is by calling eval_pattern directly from lib (alongside calling validate) -- but that's a fairly expensive operation to do twice as many times, so if someone can think of a better method I'd love to hear it.

re: Ratio, I think there probably is a way to (de)serialize it? I was pretty unfamiliar with Rust when I wrote that part and this was my workaround, but the docs suggest it should work, so I assume I overlooked something.

amyjzhu:
👍

A stopover to minimize the amount of duplicated code might be for the validation method to add to a list of counterexamples, and then write the counterexamples separately?

vee-effekt:
Oh, yeah, I could probably do something like that! Counterexamples could be added to the Analysis data, maybe?

src/bin/rational.rs
```diff
@@ -288,46 +344,110 @@ impl SynthLanguage for Math {
...
+                    println!("{} = {}", element.variable.to_string(), element.value);
+                } */
+                ces.push(ce);
+            }
+
+            // As one might expect, it's hard to avoid writing duplicate counterexamples to the file, so
+            // we get rid of them here.
+            ces.sort();
+            ces.dedup();
+            // add the ces into our randomly-generated cvecs, padding out with None where there
```
chandrakananandi, line 419: thanks for the documentation :)

src/bin/rational.rs
```diff
@@ -288,46 +344,110 @@ impl SynthLanguage for Math {
...
+            // is a counterexample with no assignment to a given variable
+            // (e.g., ?a = 1/2, ?b = 3/4 -> a: Some(1/2), b: Some(3/4), c: None)
+            //
+            // Currently, just take the first 10 counterexamples, since each run generates new
+            // counterexamples (usually) and that gets out-of-hand fast.
+            for mut ce in ces.drain(0..cmp::min(synth.params.num_ces, ces.len())) {
+                ce.sort_by(|a, b| a.variable.cmp(&b.variable));
+                for (i, (_var, cvec)) in vars.iter_mut().sorted().enumerate() {
+                    // println!("{}: {:?}, {:?}\n", i, _var, cvec);
```
chandrakananandi, line 428: remove prints

src/bin/rational.rs
```diff
@@ -288,46 +344,110 @@ impl SynthLanguage for Math {
...
+            ces.sort();
+            ces.dedup();
+            // add the ces into our randomly-generated cvecs, padding out with None where there
+            // is a counterexample with no assignment to a given variable
+            // (e.g., ?a = 1/2, ?b = 3/4 -> a: Some(1/2), b: Some(3/4), c: None)
+            //
+            // Currently, just take the first 10 counterexamples, since each run generates new
+            // counterexamples (usually) and that gets out-of-hand fast.
+            for mut ce in ces.drain(0..cmp::min(synth.params.num_ces, ces.len())) {
+                ce.sort_by(|a, b| a.variable.cmp(&b.variable));
```
chandrakananandi, line 426: why sorting?

src/bin/rational.rs
```diff
@@ -479,6 +601,33 @@ impl SynthLanguage for Math {
 
             let lvec = Self::eval_pattern(lhs, &env, n);
             let rvec = Self::eval_pattern(rhs, &env, n);
+
+            // if we are writing cvecs to a file
+            if synth.params.write_ces {
```
chandrakananandi, line 606: ohh, i see, so writing to the file is a separate feature from using the cvecs from a file. Ok this makes sense i think. You can ignore my previous comment on why we have both, but do consider renaming them to be a bit more indicative of what they are used for. Maybe `persist_cvec` is a better name for this flag

src/bin/rational.rs
```diff
@@ -479,6 +601,33 @@ impl SynthLanguage for Math {
...
+                let mut file = OpenOptions::new()
+                    .write(true)
+                    .append(true)
+                    .open("counterexamples.txt")
```
chandrakananandi, line 610: not txt, i think json right?

src/bv.rs
```diff
@@ -387,6 +387,22 @@ macro_rules! impl_bv {
 
                     let lvec = Self::eval_pattern(lhs, &env, n);
                     let rvec = Self::eval_pattern(rhs, &env, n);
+
```
chandrakananandi, line 390: remove unused code

src/lib.rs
```diff
@@ -1251,9 +1251,12 @@ pub struct SynthParams {
     ///////////////////
     #[clap(long)]
     pub prior_rules: Option<String>,
-
     #[clap(long, conflicts_with = "iters")]
     pub workload: Option<String>,
+    #[clap(long)]
```
chandrakananandi, line 1256: again, we should use better name.

src/bin/rational.rs
```diff
@@ -45,6 +55,52 @@ define_language! {
...
+        Counterexample {
+            variable: var,
+            value: val,
+        }
+    }
+}
+
+// Seems Ratio doesn't implement de/serialization, so I'm re-implementing it, basically.
+#[derive(Serialize, Deserialize, Clone)]
+struct Fraction {
```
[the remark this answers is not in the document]
chandrakananandi: ah yeah @amyjzhu is saying the same thing I believe. We don't need so many new structs. We can just use Constant as far as i can tell. Also make the var name cleaner. Right now we seem to be using String in Assignment and then Symbol in CounterExample.

src/bin/rational.rs
```diff
@@ -479,6 +601,33 @@ impl SynthLanguage for Math {
...
+                    .open("counterexamples.txt")
+                    .unwrap();
+
+                for (i, (l, r)) in lvec.iter().zip(rvec.clone()).enumerate() {
+                    // println!("{}: {} = {}", i, l.clone().unwrap(), r.clone().unwrap());
+                    if l.clone().unwrap() != r.clone().unwrap() {
+                        let mut assignments: Vec<Assignment> = vec![];
+                        for key in env.keys() {
+                            let value = env.get(key).unwrap()[i].clone().unwrap();
+                            let value_to_fraction =
```
chandrakananandi, line 619: all of this can be simplified if we only use Constant.

src/bin/rational.rs
```diff
@@ -288,46 +344,110 @@ impl SynthLanguage for Math {
...
+            consts.dedup();
+
+            let consts2 = self_product(&consts, synth.params.variables);
+            for (i, item) in consts2.iter().enumerate().take(synth.params.variables) {
+                let var = egg::Symbol::from(letter(i));
+                vars.insert(var, item.clone());
+            }
+        }
+
+        // if we are adding counterexamples to the cvec, do that here (this is the persistent cvecs
```
chandrakananandi, line 385: this is the _using_ persistent cvecs part.

chandrakananandi (changes requested):
Thanks for the PR!! Very cool work :) I have several suggestions in my comments.
On top of those, two main topics to discuss:
1. we need to figure out if there is a way to make this part of lib.rs, i.e., make the persistent cvecs more general.
2. we need to add some end-to-end tests. One way to test is to make sure that the cvecs contain what you expect in a small example. We can chat about that more.

Another thing is that the file content looks a bit odd when i write the cvecs. Maybe this can be resolved with @Amy's suggestion to use Constant and not the other structs?

```

[{"variable":"?b","value":{"numer":[1,[2]],"denom":[1,[3]]}},{"variable":"?a","value":{"numer":[-1,[90]],"denom":[1,[13]]}}]

[{"variable":"?b","value":{"numer":[1,[11]],"denom":[1,[142]]}},{"variable":"?a","value":{"numer":[1,[125]],"denom":[1,[51]]}}]

[{"variable":"?b","value":{"numer":[-1,[130]],"denom":[1,[1]]}},{"variable":"?a","value":{"numer":[-1,[189]],"denom":[1,[16]]}}]

[{"variable":"?a","value":{"numer":[-1,[24]],"denom":[1,[11]]}}]

[{"variable":"?a","value":{"numer":[1,[26]],"denom":[1,[7]]}}]

[{"variable":"?a","value":{"numer":[1,[17]],"denom":[1,[47]]}}]

[{"variable":"?a","value":{"numer":[1,[55]],"denom":[1,[62]]}}]

[{"variable":"?a","value":{"numer":[-1,[13]],"denom":[1,[2]]}}]

[{"variable":"?a","value":{"numer":[1,[52]],"denom":[1,[233]]}}]

[{"variable":"?a","value":{"numer":[1,[2]],"denom":[1,[49]]}}]

[{"variable":"?a","value":{"numer":[1,[29]],"denom":[1,[28]]}}]

[{"variable":"?a","value":{"numer":[1,[175]],"denom":[1,[22]]}}]

```

src/bin/bool.rs
```diff
@@ -226,6 +226,8 @@ mod test {
         SynthParams {
             seed: 0,
             n_samples: 2,
+            num_ces: 0,
```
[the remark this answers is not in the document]
chandrakananandi: as you will see later in my notes, the "why two params" question i had is bogus. I know now why we need two :D .

src/bin/rational.rs
chandrakananandi: also note that you dont need both write and append. If setting both to true, i think only append is needed.

chandrakananandi: More importantly this wont work unless the file exists. I believe you want `create`, not `open`

chandrakananandi:
```

                let mut file = OpenOptions::new()
                    .append(true)
                    .create(true)
                    .open("counterexamples.txt")
                    .unwrap();
```

src/bin/bigint.rs
```diff
@@ -209,6 +209,13 @@ impl SynthLanguage for Math {
 
             let lvec = Self::eval_pattern(lhs, &env, n);
             let rvec = Self::eval_pattern(rhs, &env, n);
+
```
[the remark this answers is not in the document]
vee-effekt: Oops -- removed.

src/bin/rational.rs
```diff
@@ -288,46 +344,110 @@ impl SynthLanguage for Math {
...
+                ));
+                consts.push(mk_constant(
+                    &(-i.to_bigint().unwrap()),
+                    &(1.to_bigint().unwrap()),
+                ));
+            }
+            consts.sort();
+            consts.dedup();
+
+            let consts2 = self_product(&consts, synth.params.variables);
```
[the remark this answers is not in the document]
vee-effekt: Okay!

src/bin/rational.rs
vee-effekt: Yeah, it should be .json. Does create only create the file if it does not exist already? I think we want to be able to build up results over different runs, so recreating the file each time may not be ideal.

src/bin/rational.rs
```diff
@@ -288,46 +344,110 @@ impl SynthLanguage for Math {
...
+            for (i, item) in consts2.iter().enumerate().take(synth.params.variables) {
+                let var = egg::Symbol::from(letter(i));
+                vars.insert(var, item.clone());
+            }
+        }
+
+        // if we are adding counterexamples to the cvec, do that here (this is the persistent cvecs
+        // part)
+        if synth.params.num_ces > 0 {
```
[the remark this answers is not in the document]
vee-effekt: Changed.

src/bin/str.rs
```diff
@@ -263,7 +263,7 @@ impl SynthLanguage for Lang {
         {
             let int_var = Symbol::from(letter(i + synth.params.variables));
             let id = egraph.add(Lang::Var(int_var));
-            egraph[id].data.cvec = item.iter().cycle().cloned().take(cvec_len).collect();
+            egraph[id].data.cvec = item.iter().cycle().take(cvec_len).cloned().collect();
```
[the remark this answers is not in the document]
vee-effekt: This was a Rust suggestion as Clippy was failing with the original code; not sure why.

src/bin/rational.rs
chandrakananandi: yes I believe it will do the right thing. The current version will crash because open wants the file to exist already but create doesnt. Do double check though!

src/bin/rational.rs
```diff
@@ -513,6 +559,65 @@ pub fn sampler(rng: &mut Pcg64, b1: u64, b2: u64, num_samples: usize) -> Vec<Rat
...
+            }
+            rat = rat.drain(1..).collect::<String>();
+            rat = rat.trim().to_string();
+            rat = rat.trim_matches(braces).to_string();
+            first_char = rat.chars().next().unwrap();
+        }
+        let reals: Vec<&str> = rat.split(' ').collect();
+        let mut num: f32 = reals[0].parse::<f32>().ok().unwrap();
+        num *= negative;
+        if reals.len() == 1 {
```
ajpal, line 605: maybe cleaner to do something like
`let denom = if reals.len() == 1 { 1 } else {reals[1].parse::<f32>().ok().unwrap()}`
then just
`consts.push(Constant::new(num.to_bigint().unwrap(), denom.to_bigint().unwrap()));` without needing the whole thing in the if

src/lib.rs
```diff
@@ -393,8 +401,73 @@ impl<L: SynthLanguage> Synthesizer<L> {
 
         // initialize egraph with variables and constants
         L::init_synth(&mut synth);
-        synth.initial_egraph = synth.egraph.clone();
 
+        if num_ces > 0 {
```
ajpal, line 405: you can just do `params.num_ces` here, I don't think you need to make a local variable for it

src/lib.rs
```diff
@@ -393,8 +401,73 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
+            let file = File::open(&synth.params.ces_file);
+            assert!(file.is_ok());
+            let reader = BufReader::new(file.unwrap());
+            // deserialize from file
+            let mut ces: Vec<Vec<<L as SynthLanguage>::Constant>> = vec![];
+            // let mut ce: Vec<<L as SynthLanguage>::Constant> = vec![];
```
ajpal, line 411: delete commented code

ajpal (commented):
sorry this review took a while!

src/lib.rs
vee-effekt: weird rust ownership stuff, may be a more elegant way to do this

src/bin/rational.rs
```diff
@@ -479,7 +503,29 @@ impl SynthLanguage for Math {
...
             let lvec = Self::eval_pattern(lhs, &env, n);
             let rvec = Self::eval_pattern(rhs, &env, n);
-            ValidationResult::from(lvec == rvec)
+
+            let mut assignment = vec![];
+            // if we are writing cvecs to a file
+            if synth.params.write_ces {
+                for (i, (l, r)) in lvec.iter().zip(rvec.clone()).enumerate() {
+                    if l.clone().unwrap() != r.clone().unwrap() {
+                        // Since are no longer using structs, we want to preserve an ordering for the variables.
```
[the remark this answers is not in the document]
vee-effekt: Essentially just the idea that the first element on a line corresponds to a, the next to b, etc.

vee-effekt:
This PR addresses #51.

- A user specified number of counterexamples can be deserialized from a file using the command line.
- Serializing counterexamples to a file can be toggled via the command line.
- Counterexamples are supported across all domains for fuzzing and over the rational domain for SMT.
- If a counterexample is "used" -- that is, a cvec that would otherwise not have flagged an unsound candidate rule now does so because of the inclusion of a counterexample, that information is logged.

PR #60, opened 2022-05-10
fixing derive maybe?

chandrakananandi:
I think derive should only ever allow derivation of `All vs All` or `New vs New`. I was unable to come up with a scenario where we should allow `All vs New` or `New vs All`. This PR adds this restricts. Previously, it would technically be possible to do these potentially incorrect comparisons.


bksaiki (approved):
I think this change is fine. I believe the command line flags restrict the behavior of the derive tool to All vs All or New vs New anyway. In fact, the RuleType enum might be unnecessary.

chandrakananandi:
> I think this change is fine. I believe the command line flags restrict the behavior of the derive tool to All vs All or New vs New anyway. In fact, the RuleType enum might be unnecessary.

@bksaiki how so? which CL flags?

bksaiki:
I believe `params.new_eqs`, but I could be wrong.

chandrakananandi:
> I believe `params.new_eqs`, but I could be wrong.

you are right, i just fixed. Pushing.

chandrakananandi:
> > I believe `params.new_eqs`, but I could be wrong.
> 
> you are right, i just fixed. Pushing.

@bksaiki  let me know if this is better.

bksaiki (approved):
Looks good!

PR #61, opened 2022-06-09
Add Nat domain

ajpal:
To discuss: I had to change the signature of `mk_constant` to take in the egraph to make it possible to recursively construct the Nat term. Is this okay? I couldn't think of another way to do it, but it does require changing the signature across all domains so not sure if it's a change we want to make.

With a corresponding workload of 5 atom terms, finds these rules:
```
(* ?c (* ?b ?a)) ==> (* ?a (* ?b ?c))
(+ ?c (+ ?b ?a)) ==> (+ ?a (+ ?b ?c))
(* ?b ?a) ==> (* ?a ?b)
(+ ?b ?a) ==> (+ ?a ?b)
(+ ?b (S ?a)) ==> (+ ?a (S ?b))
(S (+ ?b ?a)) <=> (+ ?a (S ?b))
(+ ?a (* ?b ?a)) <=> (* ?a (S ?b))
?a <=> (+ ?a Z)
?a <=> (* ?a (S Z))
(* ?a Z) ==> Z
```

### DIFF
--- a/src/bin/bigint.rs
+++ b/src/bin/bigint.rs
@@ -70,7 +70,7 @@ impl SynthLanguage for Math {
         }
     }
 
-    fn mk_constant(c: Self::Constant, _egraph: &EGraph<Self, SynthAnalysis>) -> Self {
+    fn mk_constant(c: Self::Constant, _egraph: &mut EGraph<Self, SynthAnalysis>) -> Self {
         Math::Num(c)
     }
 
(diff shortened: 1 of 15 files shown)


ajpal:
closed in favor of https://github.com/uwplse/ruler/pull/71

chandrakananandi:
I know this is closed, but just wondering if @bksaiki or @ztatlock have thoughts on this design question. I personally think it's fine as long as we have comments to explain why we needed it.

PR #62, opened 2022-07-15
is_constant instead of to_constant

ajpal:
I think `to_constant` is only used by `is_constant` so we can probably just implement `is_constant` in each domain directly. This seems like a harmless little cleanup, but please let me know if there's some context I'm missing!

### DIFF
--- a/src/bin/bigint.rs
+++ b/src/bin/bigint.rs
@@ -62,12 +62,8 @@ impl SynthLanguage for Math {
         Math::Var(sym)
     }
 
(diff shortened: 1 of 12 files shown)


chandrakananandi (approved):
Thanks for this cleanup! `to_constant` was an artifact of having all the "constructors" and "convertors" in place, like we have `to_var` , `mk_var`, `mk_constant` etc. Since we don't seem to be using `to_constant` anywhere, this is a good idea.

PR #65, opened 2022-08-09
Switch pred to use rats not usize

ajpal:
This is necessary to add the math operators to the language without having to deal with overflow

### DIFF
--- a/src/bin/pred.rs
+++ b/src/bin/pred.rs
@@ -227,7 +227,7 @@ impl SynthLanguage for Pred {
     fn init_synth(synth: &mut Synthesizer<Self>) {
         let cvec_len = 300;
         let mut egraph: EGraph<Pred, SynthAnalysis> = EGraph::new(SynthAnalysis {
-            cvec_len: cvec_len,
+            cvec_len,
             constant_fold: ConstantFoldMethod::CvecMatching,
             rule_lifting: false,
         });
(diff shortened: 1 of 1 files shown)


src/bin/pred.rs
```diff
@@ -223,24 +225,31 @@ impl SynthLanguage for Pred {
     }
 
     fn init_synth(synth: &mut Synthesizer<Self>) {
+        let cvec_len = 300;
```
chandrakananandi, line 228: this is because too many unsound rules if cvec is small?

src/bin/pred.rs
```diff
@@ -223,24 +225,31 @@ impl SynthLanguage for Pred {
...
+            constant_fold: ConstantFoldMethod::CvecMatching,
             rule_lifting: false,
         });
 
         for i in 0..synth.params.variables {
             let rng = &mut synth.rng;
-            let mut i_vals = vec![];
+
+            let mut samples = vec![];
+            samples.append(&mut sampler(rng, 8, 6, cvec_len / 3));
```
chandrakananandi, line 239: hopefully with the SMT based validation, this won't be needed!

chandrakananandi (approved):

src/bin/pred.rs
ajpal: 300 was chosen semi-arbitrarily. Other domains have cvecs with hundreds to a couple-thousand elements, so 300 seemed like in the right ballpark.

PR #67, opened 2022-08-10
Fix String domain

ajpal:
String was broken on `main`.
I tried to root-cause the issue for a bit but couldn't quite tell what was going on.
Instead, I re-implemented it, and used it as an opportunity to switch from `SmallVec<u8>` to `SmallString`. I also added fuzzing for rule validation, which was at least part of the problem.

Here are the rules I find with [this recipe](https://github.com/uwplse/recipes/blob/str2/str/str.spec)
```
(+ ?nb ?na) <=> (+ ?na ?nb)
(str.replace ?sb ?sa ?sa) => ?sb
(str.replace ?sb ?sb ?sa) => ?sa
(str.replace ?sb "" ?sa) <=> (str.++ ?sa ?sb)
(str.indexof ?sb "" ?na) => (- ?na ?na)
(str.substr ?sb ?na 1) <=> (str.at ?sb ?na)
(str.substr "" ?nb ?na) => ""
(str.substr ?sb ?na 0) => ""
(str.indexof "" ?sb ?na) => (str.indexof ?sb ?sb 1)
(str.to_int ?sa) => -1
?na <=> (- ?na 0)
(- ?na ?na) => 0
?na <=> (+ 0 ?na)
(str.indexof ?sa ?sa 0) => 0
(str.replace "" ?sa "") => ""
(str.substr "a" ?na ?na) => ""
(str.substr "a" 1 ?na) => ""
(str.replace "" "a" ?sa) => ""
(str.indexof ?sa ?sa 1) <=> (str.indexof "a" ?sa 1)
```

And here are the enumerate (iter 1) rules:
```
(+ ?nb ?na) <=> (+ ?na ?nb)
(str.replace ?sb ?sb ?sa) => ?sa
(str.replace ?sb ?sa ?sa) => ?sb
(str.substr ?sb ?na 1) <=> (str.at ?sb ?na)
(str.substr ?sb ?na 0) => (str.at ?sb -1)
(str.substr ?sb -1 ?na) => (str.at ?sb -1)
(str.indexof ?sb ?sa -1) => (str.indexof ?sa ?sa -1)
(str.to_int ?sa) => -1
?na <=> (- ?na 0)
?na <=> (+ 0 ?na)
(- ?na ?na) => 0
(- ?na 1) <=> (+ -1 ?na)
(- ?na -1) <=> (+ 1 ?na)
(str.indexof ?sa ?sa 0) => 0
?sa <=> (str.substr ?sa 0 -1)
(str.indexof ?sa ?sa 1) <=> (str.indexof ?sa ?sa -1)
(str.substr ?sa 1 -1) <=> (str.at ?sa -1)
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ z3 = "0.10.0"
 rustc-hash = "1"
 symbolic_expressions = "5"
 rayon = "1"
-smallvec = { version = "1.6", features = ["union", "const_generics"] }
+smallstr = "0.3"
 
 [dependencies.egg]
 version = "0.8"
--- a/src/bin/str.rs
+++ b/src/bin/str.rs
@@ -5,44 +5,61 @@ Experimental String domain.
 use egg::*;
 use ruler::*;
 
-use std::ops::*;
-
-use smallvec::SmallVec;
 use std::fmt;
 
-pub type SmallStr = SmallVec<[u8; 24]>;
+use rand::{seq::SliceRandom, Rng};
+use rand_pcg::Pcg64;
+use smallstr::SmallString;
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct SVar(egg::Symbol);
+impl fmt::Display for SVar {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "s")
+    }
+}
+impl std::str::FromStr for SVar {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.starts_with('s') {
+            Ok(SVar(Symbol::from(s)))
+        } else {
+            Err(())
+        }
+    }
+}
+
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
+pub struct NVar(egg::Symbol);
+impl fmt::Display for NVar {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "n")
+    }
+}
+impl std::str::FromStr for NVar {
+    type Err = ();
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        if s.starts_with('n') {
+            Ok(NVar(Symbol::from(s)))
+        } else {
+            Err(())
+        }
+    }
+}
+
 #[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-/// Two types of Constants: Num and Str.
 pub enum Constant {
     Num(i64),
-    Str(SmallStr),
+    Str(SmallString<[u8; 24]>),
 }
-
-impl Constant {
-    fn to_num(&self) -> Option<i64> {
-        match self {
-            Constant::Num(n) => Some(*n),
-            Constant::Str(_) => None,
-        }
-    }
-    fn to_slice(&self) -> Option<&[u8]> {
-        match self {
-            Constant::Num(_) => None,
-            Constant::Str(s) => Some(s.as_slice()),
-        }
-    }
-    fn to_str(&self) -> Option<&str> {
-        self.to_slice()
-            .map(|bytes| unsafe { std::str::from_utf8_unchecked(bytes) })
-    }
-}
-
 impl fmt::Display for Constant {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         match self {
             Constant::Num(n) => n.fmt(f),
             Constant::Str(s) => {
-                let s = std::str::from_utf8(s).unwrap();
+                let s = s.as_str();
                 write!(f, "\"")?;
                 s.fmt(f)?;
                 write!(f, "\"")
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -26,7 +26,7 @@ z3 = "0.10.0"
 rustc-hash = "1"
 symbolic_expressions = "5"
 rayon = "1"
-smallvec = { version = "1.6", features = ["union", "const_generics"] }
+smallstr = "0.3"
```
chandrakananandi, line 29: what's the advantage of smallstr over smallvec again?

src/bin/str.rs
```diff
@@ -54,30 +71,56 @@ impl fmt::Display for Constant {
...
-        match i64::from_str(s) {
-            Ok(i) => Ok(Self::Num(i)),
-            Err(_) => {
-                let s = s.trim_matches('"');
-                Ok(Self::Str(s.as_bytes().into()))
-            }
+        if let Ok(i) = i64::from_str(s) {
+            Ok(Self::Num(i))
+        } else {
+            Ok(Self::Str(SmallString::from_str(s.trim_matches('"'))))
```
chandrakananandi, line 77: why is the `trim_matches` needed?

src/bin/str.rs
```diff
@@ -89,88 +132,92 @@ define_language! {
...
+                Constant::Num(_) => Type::Number,
+                Constant::Str(_) => Type::String,
+            },
+            Lang::SVar(_)
+            | Lang::Concat(_)
+            | Lang::Replace(_)
+            | Lang::At(_)
+            | Lang::SubStr(_)
+            | Lang::FromInt(_) => Type::String,
+            Lang::NVar(_)
```
chandrakananandi, line 154: ah yeah `str` is an interesting domain because `get_type` is not all the same for the ops! Cool cool.

src/bin/str.rs
```diff
@@ -89,88 +132,92 @@ define_language! {
...
+                Some(Constant::Str(SmallString::from_str(&out)))
             }),
             Lang::At([a, i]) => map!(v, a, i => {
-                let a = a.to_slice().unwrap();
+                let a = a.to_smallstr().unwrap();
                 let i = i.to_num().unwrap();
-                if 0 <= i && i < (a.len() as _) {
-                    let i = i as usize;
-                    Some(Constant::Str(a[i..i+1].into()))
+                if i < 0 || i > (a.len() as _) {
```
chandrakananandi, line 184: shouldn't this be `i >= (a.len() as _)` ?

src/bin/str.rs
```diff
@@ -54,30 +71,56 @@ impl fmt::Display for Constant {
...
+        if let Ok(i) = i64::from_str(s) {
+            Ok(Self::Num(i))
+        } else {
+            Ok(Self::Str(SmallString::from_str(s.trim_matches('"'))))
+        }
+    }
+}
+
+impl Constant {
+    fn to_smallstr(&self) -> Option<SmallString<[u8; 24]>> {
```
chandrakananandi, line 83: why do we need both to_str and to_smallstr?

src/bin/str.rs
```diff
@@ -89,88 +132,92 @@ define_language! {
...
-                    let i = i as usize;
-                    Some(Constant::Str(a[i..j].into()))
-                } else {
-                    Some(Constant::Str(SmallStr::default()))
+                let a = a.to_smallstr().unwrap();
+                let n = a.len() as usize;
+                let i = i.to_num().unwrap() as usize;
+                let len = len.to_num().unwrap() as usize;
+                let j = (i + len).min(n) as usize;
+                match a.get(i..j) {
```
chandrakananandi, line 197: do we not need the condition  `0 <= i && i < n && len > 0`? Or does the `a.get(i..j)` do the right thing here?

src/bin/str.rs
```diff
@@ -89,88 +132,92 @@ define_language! {
...
-                if 0 <= i && i <= (s.len() as _) {
-                    let i = i as usize;
-                    match s[i..].find(t) {
-                        None => Some(Constant::Num(-1)),
-                        Some(j) => Some(Constant::Num(j as _))
-                    }
-                } else {
-                    Some(Constant::Num(-1))
+                let i = i.to_num().unwrap() as usize;
+                match s.get(i..s.len()).unwrap_or("").find(t) {
```
chandrakananandi, line 218: i have the same question here as I had for `SubStr` but I think the `match` probably does the right thing here too? It will just be `None` if the bounds are not right?

chandrakananandi (changes requested):
Thanks for fixing this. Looks great, I just have a couple of questions and for `Lang::At` we should double check the bound in the `if`.

Cargo.toml
ajpal: I think it leads to clearer code than using a smallvec of bytes since the data type is actually meant for strings. It also allows us to get rid of this [unsafe code block](https://github.com/uwplse/ruler/blob/main/src/bin/str.rs#L36)

@mwillsey can probably give a more detailed answer though!

mwillsey: Yeah, smallstr makes sense to me.

src/bin/str.rs
ajpal: [get](https://doc.rust-lang.org/std/string/struct.String.html#method.get) should just return `None` if the indices aren't valid, so I don't think we need to check.

src/bin/str.rs
ajpal: Yeah, if the bounds are not right, the `get()` will return `None`, and then `None.unwrap_or("")` will be `""`, and then you call `find` on that (which will also return `None`)

src/bin/str.rs
ajpal: that was in the original implementation ([here](https://github.com/uwplse/ruler/blob/main/src/bin/str.rs#L60)) so I just left it. I think it just removes leading/trailing quotes?

src/bin/str.rs
ajpal: Probably simpler to just have to_str

src/bin/str.rs
ajpal: Yep, you're right, my bad!

chandrakananandi (approved):

PR #69, opened 2022-08-12
Pred verify

chandrakananandi:
This PR has the following changes:

- using z3 to validate rules
- using the important-cvec-offset approach for initializing cvecs for `NVar`, similar to the rational domain.
- `le -> lt`, `ge -> gt`
- an entry for `pred` in the config

### DIFF
--- a/src/bin/pred.rs
+++ b/src/bin/pred.rs
@@ -5,11 +5,11 @@ use std::{
 };
 
 use egg::*;
-use num::{rational::Ratio, Zero};
 use num::{
     bigint::{BigInt, RandBigInt, Sign, ToBigInt},
     ToPrimitive,
 };
+use num::{rational::Ratio, Zero};
 use rand::{seq::SliceRandom, Rng};
 use rand_pcg::Pcg64;
 use ruler::*;
@@ -238,18 +238,17 @@ impl SynthLanguage for Pred {
     }
 
     fn init_synth(synth: &mut Synthesizer<Self>) {
-
         let rng = &mut synth.rng;
         let mut consts: Vec<Option<Constant>> = vec![];
 
         for i in 0..synth.params.important_cvec_offsets {
             consts.push(mk_constant(
                 &i.to_bigint().unwrap(),
-                &((i as u32).to_bigint().unwrap()),
+                &((1_u32).to_bigint().unwrap()),
             ));
             consts.push(mk_constant(
                 &(-i.to_bigint().unwrap()),
-                &((1 as u32).to_bigint().unwrap()),
+                &((1_u32).to_bigint().unwrap()),
             ));
         }
 
@@ -307,7 +306,7 @@ impl SynthLanguage for Pred {
                 SatResult::Sat => {
                     println!("z3 validation: failed for {} => {}", lhs, rhs);
                     ValidationResult::Invalid
-                },
+                }
                 SatResult::Unknown => {
                     synth.smt_unknown += 1;
                     ValidationResult::Unknown
--- a/src/bin/pred.rs
+++ b/src/bin/pred.rs
@@ -266,7 +266,9 @@ impl SynthLanguage for Pred {
             let lexpr = egg_to_z3(&ctx, Self::instantiate(lhs).as_ref());
             let rexpr = egg_to_z3(&ctx, Self::instantiate(rhs).as_ref());
             match (lexpr, rexpr) {
-                (Z3::Z3Bool(lb), Z3::Z3Bool(rb)) => solver.assert(&z3::ast::Bool::not(&lb._eq(&rb))),
+                (Z3::Z3Bool(lb), Z3::Z3Bool(rb)) => {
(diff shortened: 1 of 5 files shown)


src/bin/pred.rs
```diff
@@ -73,6 +77,15 @@ define_language! {
   }
 }
 
+/// Return a non-zero constant.
+fn mk_constant(n: &BigInt, d: &BigInt) -> Option<Constant> {
```
ajpal, line 81: Why have this return an Option? Seems simpler to just panic if `d` is zero, since we know it is only called with `d` being `1`.

src/bin/pred.rs
```diff
@@ -73,6 +77,15 @@ define_language! {
   }
 }
 
+/// Return a non-zero constant.
```
ajpal, line 80: nit: technically could be 0 right? We don't check that `n` is non-zero

src/bin/pred.rs
```diff
@@ -225,34 +238,40 @@ impl SynthLanguage for Pred {
...
-            let n_vals: Vec<Option<Constant>> = samples.iter().map(|x| Some(x.clone())).collect();
-            let mut b_vals = vec![];
+        for (i, n_vals) in cs.iter().enumerate().take(synth.params.variables) {
             let n_id = egraph.add(Pred::NVar(NVar(Symbol::from("n".to_owned() + letter(i)))));
             let b_id = egraph.add(Pred::BVar(BVar(Symbol::from("b".to_owned() + letter(i)))));
+            let mut b_vals = vec![];
             for _ in 0..cvec_len {
-                b_vals.push(Some(Constant::Bool(rng.gen::<bool>())));
+                b_vals.push(Some(Constant::Bool(synth.rng.gen::<bool>())));
             }
```
ajpal, line 271: Maybe more readable to do this in two loops?
```
for (i, n_vals) in cs.iter().enumerate().take(synth.params.variables) {
    let n_id = egraph.add(Pred::NVar(NVar(Symbol::from("n".to_owned() + letter(i)))));
    egraph[n_id].data.cvec = n_vals.clone();
}
for _ in 0..synth.params.variables {
    let b_id = egraph.add(Pred::BVar(BVar(Symbol::from("b".to_owned() + letter(i)))));
    let mut b_vals = vec![];
    for _ in 0..cvec_len {
        b_vals.push(Some(Constant::Bool(synth.rng.gen::<bool>())));
    }
    egraph[b_id].data.cvec = b_vals.clone();
}
```

src/bin/pred.rs
```diff
@@ -323,6 +370,111 @@ pub fn sampler(rng: &mut Pcg64, num: u64, denom: u64, num_samples: usize) -> Vec
...
+
+impl<'a> Z3<'a> {
+    fn get_z3bool(&self) -> Option<z3::ast::Bool<'a>> {
+        match self {
+            Z3::Z3Bool(b) => Some(b.clone()),
+            Z3::Z3Real(_) => None,
+        }
+    }
+
+    fn get_z3rat(&self) -> Option<z3::ast::Real<'a>> {
```
ajpal, line 387: nit: `get_z3real`?

src/bin/pred.rs
```diff
@@ -323,6 +370,111 @@ pub fn sampler(rng: &mut Pcg64, num: u64, denom: u64, num_samples: usize) -> Vec
...
+                ))),
+            },
+            Pred::BVar(bv) => buf.push(Z3::Z3Bool(z3::ast::Bool::new_const(ctx, bv.to_string()))),
+            Pred::NVar(nv) => buf.push(Z3::Z3Real(z3::ast::Real::new_const(ctx, nv.to_string()))),
+            Pred::Lt([a, b]) => pushbuf!(buf, a, b, lt, get_z3rat, Real),
+            Pred::Leq([a, b]) => pushbuf!(buf, a, b, le, get_z3rat, Real),
+            Pred::Gt([a, b]) => pushbuf!(buf, a, b, gt, get_z3rat, Real),
+            Pred::Geq([a, b]) => pushbuf!(buf, a, b, ge, get_z3rat, Real),
+            Pred::And([a, b]) => pushbuf!(buf, ctx, a, b, and, get_z3bool, Bool),
+            Pred::Or([a, b]) => pushbuf!(buf, ctx, a, b, or, get_z3bool, Bool),
```
ajpal, line 438: Why do these need `ctx` and the rest don't?

ajpal (approved):
left a few questions / small comments but lgtm!

src/bin/pred.rs
chandrakananandi: I am not really sure. It's just how the z3 rust api is built. Some of them don't pass the context around 🤷‍♀️

src/bin/pred.rs
chandrakananandi: yeah I guess we can here, in the future if we end up using it more generally, can change it back to Option.

src/bin/pred.rs
chandrakananandi: good catchm I meant denominator, will fix.

PR #70, opened 2022-08-17
Add -> op to bool and pred

ajpal:


chandrakananandi (approved):
Fantastic, thanks so much for doing this so quickly! 🙏 

I guess it does cause CI regression and unit tests to fail because now the ruleset has changed. Should we update the rest or just not enumerate `implies` by default?

PR #72, opened 2022-09-27
[Bug fix] Return empty vec if lhs & rhs are already unioned

ajpal:
This was preventing egg from ever saturating!
Basically, we need to return an empty vec if applying the rewrite didn't actually "do anything"
See https://github.com/egraphs-good/egg/blob/main/src/run.rs#L560 and https://github.com/egraphs-good/egg/blob/main/src/run.rs#L589 for how this ends up getting used.


chandrakananandi (approved):
Wow! great find! Thanks for the fix. :shipit:

mwillsey (approved):
Looks good! Nice catch.

PR #73, opened 2022-09-30
Rule lifting rewrite

ajpal:
One regression test is failing due to not being able to derive `(xor ?a (or ?b ?a)) <=> (and ?b (not ?a))`
Still not exactly sure how/why we're losing that proving power. I'm looking into it, but I feel like it's still worth starting code review now

### DIFF
--- a/src/bin/trig.rs
+++ b/src/bin/trig.rs
@@ -232,29 +232,10 @@ impl SynthLanguage for Math {
     }
 
     fn is_extractable(&self) -> bool {
-        matches!(
+        !matches!(
             self,
-            // Standard extractable nodes
-            Math::Sin(_)
-                | Math::Cos(_)
-                | Math::Tan(_)
-                | Math::Neg(_)
-                | Math::Add(_)
-                | Math::Sub(_)
-                | Math::Mul(_)
-                | Math::Div(_)
-                | Math::RealConst(_)
-                | Math::Pi
-                | Math::Var(_)
-
-                // Interesting complex nodes
-                // | Math::Cis(_)
-                // | Math::Imag
-
-                // Reciprocal trig
-                | Math::Csc(_)
-                | Math::Sec(_)
-                | Math::Cot(_)
+            Math::Imag |
+            Math::Cis(_),
         )
     }
 
--- a/src/bin/trig.rs
+++ b/src/bin/trig.rs
@@ -259,17 +259,7 @@ impl SynthLanguage for Math {
     }
 
     fn init_synth(synth: &mut Synthesizer<Self>) {
-        // disabled operators (TODO: validate input)
-        let disabled_ops: Vec<&str> = if let Some(s) = &synth.params.disabled_ops {
-            s.split(' ').collect()
-        } else {
-            vec![]
-        };
-
-        // predicate if disabled
-        let allowedp = |s: &str| !disabled_ops.iter().any(|x| x.eq(&s));
-
-        let mut egraph = EGraph::new(SynthAnalysis {
+        let egraph = EGraph::new(SynthAnalysis {
             cvec_len: 0,
             constant_fold: if synth.params.no_constant_fold {
                 ConstantFoldMethod::NoFold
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -891,290 +869,166 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
+        );
 
+        let mut candidates: EqualityMap<L> = EqualityMap::default();
         for ids in found_unions.values() {
-            for win in ids.windows(2) {
-                if self.params.keep_all
-                    || (self.egraph[win[0]].nodes.iter().any(L::is_extractable)
-                        && self.egraph[win[1]].nodes.iter().any(L::is_extractable))
-                {
+            for id1 in ids.clone() {
```
bksaiki, line 980: Is there a reason for this method of iteration rather than a sliding window of 2?

src/lib.rs
```diff
@@ -891,290 +869,166 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
                                 }
                             }
                         }
                     }
                 }
-
-                self.egraph.union(win[0], win[1]);
             }
         }
+        candidates.retain(|_, v| L::is_allowed_rewrite(&v.lhs, &v.rhs));
```
bksaiki, line 1026: So much cleaner!

bksaiki (commented):
Looks much better. Possible things that need looking into
  (1) strange loss of derivability for one test
  ~~(2) effects on trig domain since the scheduler was changed
      (wouldn't be surprised if doesn't come back)~~
 (2) effects on rule lifting domains since the scheduler was changed.
    It seems the bv2 from bool domain went from 2 sec -> 87 sec in CI.
    I think it's safe to say trig won't come back.
    Scheduling is still a problem.

src/lib.rs
```diff
@@ -448,25 +448,11 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
-                } else {
-                    Ok(())
-                }
-            })
-            .with_node_limit(self.params.eqsat_node_limit * 2)
+            .with_node_limit(self.params.eqsat_node_limit)
             .with_iter_limit(self.params.eqsat_iter_limit)
             .with_time_limit(Duration::from_secs(self.params.eqsat_time_limit))
-            .with_scheduler(scheduler)
+            .with_scheduler(SimpleScheduler)
```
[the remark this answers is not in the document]
bksaiki: See note below on effects in rule lifting domain. Moving to `SimpleScheduler` is questionable.

bksaiki:
Further investigation suggests that scheduling is not the problem (91 seconds for Backoff, 95 seconds for SimpleScheduler). Maybe the new unioning scheme?

src/lib.rs
bksaiki: I think a sliding window is better. This is causing the performance blow up. Is there some factor I'm missing?

ajpal: There was a discussion about this in [this slack thread](https://uw-cse.slack.com/archives/C01749FQE4U/p1662153465032329)
the gist is- by using a sliding window and unioning as you go, you only see a subset of potential candidates in an arbitrary/unpredictable order

src/bin/trig.rs
```diff
@@ -232,44 +232,11 @@ impl SynthLanguage for Math {
...
-                // Interesting complex nodes
-                // | Math::Cis(_)
-                // | Math::Imag
-
-                // Reciprocal trig
-                | Math::Csc(_)
-                | Math::Sec(_)
-                | Math::Cot(_)
-        )
+        !matches!(self, Math::Imag | Math::Cis(_),)
```
chandrakananandi, line 235: great!

src/lib.rs
```diff
@@ -891,290 +869,176 @@ impl<L: SynthLanguage> Synthesizer<L> {
...
+            old_eqs,
+            smt_unknown: self.smt_unknown,
+        }
+    }
 
-        let mut old_allowed_eqs: EqualityMap<L> = EqualityMap::default();
-        let mut old_forbidden_eqs: EqualityMap<L> = EqualityMap::default();
-        for (name, eq) in self.old_eqs.clone() {
+    fn run_chunk_rule_lifting(&mut self, chunk: &[RecExpr<L>]) {
+        log::info!("Adding {} terms to egraph", chunk.len());
```
chandrakananandi, line 929: is this `step 1. Add workload terms to egraph ` from the ruler doc? If so, can we add a comment?

chandrakananandi (approved):
I took a quick look at the PR and matched it with the algorithm we wrote in the ruler doc. It seems great to me! Left a suggestion for a comment but rest is good :shipit:

PR #74, opened 2022-10-14
Clean up mkrunner

ajpal:
A couple little things to clean up here:
1. I don't think we need the `size > node_limit` check (correct me if I'm wrong?)
2. We weren't actually using the node limit or time limit supplied by the command line args


chandrakananandi (approved):
I can't remember why we had this size > node_limit clause. I support removing it if it's not  improving anything!

PR #76, opened 2022-11-02
fix is_allowed, use ExtractableAstSize

ajpal:


chandrakananandi (approved):
:shipit:

src/lib.rs
```diff
@@ -1004,26 +1007,30 @@ impl<L: SynthLanguage> Synthesizer<L> {
         }
         self.egraph = new_egraph;
 
-        // 5. Run forbidden + lifting rules
+        // 5. Run all rules
```
bksaiki, line 1010: What's the argument for all vs. forbidden?

ajpal: we changed the definition of allowed to be everything that doesn't contain Cis or i (previously, it had to have trip ops to be considered allowed). So now, way more rules are classified as allowed (all the pure rational rules). We still need merges from these rules to be treated as merge candidates after we run the denotation rules.

Basically, before we run the denotation rules, we run the allowed rules to compress the egraph as much as possible.
Then we run the denotation rules to saturation.
Then we run all of the rules we have, and everything is a merge candidate because we want to find alternate ways to prove the merges that don't go through the denotation rules.

(this is in the "language-reduction" view of rule lifting, rather than thinking about "high-level" and "low-level" domains. think instead of rule lifting as "rule laundering"- we're trying to delete the forbidden rules without losing their proving power by finding more allowed rules that recoup the proving power of the forbidden rules)

PR #78, opened 2022-11-05
Parse unidirectional rules properly

ajpal:
See https://uw-cse.slack.com/archives/C01749FQE4U/p1667618113237309 for context


src/equality.rs
```diff
@@ -26,23 +26,23 @@ impl FromStr for SerializedEq {
     type Err = String;
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        if let Some((l, r)) = s.split_once("=>") {
+        if let Some((l, r)) = s.split_once("<=>") {
```
ajpal, line 29: have to do the `<=>` case before the `=>` case otherwise everything will be parsed as `=>`

bksaiki: Hilarious

src/bin/bool.rs
```diff
@@ -255,10 +255,10 @@ mod test {
     fn get_params(num_iters: usize) -> SynthParams {
         SynthParams {
             seed: 0,
-            n_samples: 2,
+            n_samples: 0,
```
bksaiki, line 258: What's with all these changes?

bksaiki (approved):
Seems good to me.

src/bin/bool.rs
ajpal: these parameters just don't match the defaults for some reason, so I changed them here so that the test is the same configuration as `cargo bool --iters 1/2`

bksaiki: Got it.

PR #83, opened 2022-11-14
Rewrite

ajpal:


src/derive.rs
```diff
@@ -0,0 +1,143 @@
...
+            }
+        })
+}
+
+pub fn derive<L: SynthLanguage>(params: DeriveParams) {
+    let test_rules = parse::<L>(&params.in1);
+    let num_test = test_rules.len();
+    let baseline_rules = parse::<L>(&params.in2);
+
+    println!("Using {} to derive {}", params.in1, params.in2);
```
chandrakananandi, line 54: don't we want to do both directions?

src/derive.rs
```diff
@@ -0,0 +1,143 @@
...
+
+            l_id = runner.egraph.find(runner.roots[0]);
+            r_id = runner.egraph.find(runner.roots[1]);
+
+            if l_id == r_id {
+                break;
+            }
+
+            // Other
+            runner = mk_runner(runner.egraph, &l, &r)
```
chandrakananandi, line 100: so just 1 round here is enough? I haven't thought this through yet, but for derivability, do we need to do a few more?

src/language.rs
```diff
@@ -129,6 +142,20 @@ pub trait SynthLanguage: egg::Language + Send + Sync + Display + FromOp + 'stati
...
+        let pattern_is_extractable = |pat: &Pattern<Self>| {
+            pat.ast.as_ref().iter().all(|n| match n {
+                ENodeOrVar::ENode(n) => n.is_extractable(),
+                ENodeOrVar::Var(_) => true,
+            })
+        };
+        pattern_is_extractable(lhs) && pattern_is_extractable(rhs)
+    }
+
+    fn is_extractable(&self) -> bool {
```
chandrakananandi, line 155: assuming rule-lifting related parts of the code are not for review for now.

src/synth.rs
```diff
@@ -0,0 +1,422 @@
...
+                    self.egraph.union(first, *id);
+                }
+            }
+        }
+        self.egraph.rebuild();
+    }
+
+    fn cvec_match(&self) -> EqualityMap<L> {
+        // cvecs [𝑎1, . . . , 𝑎𝑛] and [𝑏1, . . . , 𝑏𝑛] match iff:
+        // ∀𝑖. 𝑎𝑖 = 𝑏𝑖 ∨ 𝑎𝑖 = null ∨ 𝑏𝑖 = null and ∃𝑖. 𝑎𝑖 = 𝑏𝑖 ∧ 𝑎𝑖 ≠ null ∧ 𝑏𝑖 ≠ null
```
chandrakananandi, line 133: wow this logical formula rendering is so cool.

src/synth.rs
```diff
@@ -0,0 +1,422 @@
...
+                    if let Some(eq) = Equality::new(&e2, &e1) {
+                        candidates.insert(eq.name.clone(), eq);
+                    }
+                }
+            }
+        }
+        candidates
+    }
+
+    // TODO: Figure out what to do with this- it doesn't match the definition
```
chandrakananandi, line 177: IMHO this is fine. It definitely makes our tool incomplete but it is not totally unreasonable. Ruler is a synthesis tool, and such tools always have to work around state space explosion problems.

src/synth.rs
```diff
@@ -0,0 +1,422 @@
...
+                            candidates.insert(eq.name.clone(), eq);
+                        }
+                    }
+                }
+            }
+        }
+        candidates
+    }
+
+    fn run_rule_lifting(&mut self) -> EqualityMap<L> {
```
chandrakananandi, line 320: so we should make sure this matches what we discussed, or maybe we need to go over this again and compare this version against the ones we drafted in the paper.

chandrakananandi (commented):
This is a massive cleanup. Left some comments, not all of which may apply after the draft is complete. Thanks for doing this!

tests/rat-ema.json
```diff
@@ -1,393 +0,0 @@
-{
-  "params": {
-    "seed": 0,
```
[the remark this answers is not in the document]
ajpal: Ema isn't part of ruler anymore (it'll all be taken care of by enumo workloads)

src/bin/bv4.rs
```diff
@@ -0,0 +1,10 @@
+/*!
```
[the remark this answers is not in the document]
ajpal: Yep

src/derive.rs
```diff
@@ -0,0 +1,143 @@
...
+        "Partitioned {} eqs into {} sat and {} other",
+        num_test,
+        sat.len(),
+        other.len()
+    );
+
+    let mut derivable = vec![];
+    let mut not_derivable = vec![];
+
+    baseline_rules.into_iter().for_each(|eq| {
```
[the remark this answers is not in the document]
ajpal: Yeah, I can add it back (but might do it as a follow-on just to keep this diff smaller). My general strategy is to do the simplest thing first and only add complexity when I run into issues with the simple thing.

src/derive.rs
ajpal: I don't think so- I think you should run it both ways if you want both ways, but there are legitimate use-cases (like Halide) where you want to know how many X rules can be derived by Y but you don't care about the reverse

src/derive.rs
ajpal: see line 84- there's an outer loop for num_iters

src/language.rs
ajpal: yeah, not quite

src/synth.rs
ajpal: I agree! I didn't expect copy/pasting from the paper to work, but it did!

src/synth.rs
ajpal: This matches the version on main currently, which is the new version from the meeting notes a few weeks ago

PR #85, opened 2022-11-22
Simplify loop

ajpal:
This shouldn't change anything functionally, it's just a more readable version of the loop

Two things:
1. We don't use the cost or the id, so we should just have a vector of the `RecExpr`s, not the tuples in `terms_ids`
2. We want to loop over the [upper triangular matrix](https://en.wikipedia.org/wiki/Triangular_matrix). The old way did this by removing the first element before the inner loop (`let ((_, e1), _) = terms_ids.remove(0);`). I think it's much more readable to structure the nested loop as
```
for i in 0..len {
  for j in i..len {
    // do something
  }
}
```

### DIFF
--- a/src/derive.rs
+++ b/src/derive.rs
@@ -108,6 +108,12 @@ pub fn derive<L: SynthLanguage>(params: DeriveParams) {
                 break;
             }
         }
+        // One more sat
+        runner = mk_runner(runner.egraph, &l, &r)
+            .with_node_limit(usize::MAX)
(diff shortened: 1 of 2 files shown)


bksaiki:
Can the loose `.z3-trace` file be nixed on this branch? It snuck in on the rewrite PR. Possibly gitignored if this is always generated.

chandrakananandi (approved):
NOICE

PR #88, opened 2022-12-02
Enumo

ajpal:


src/language.rs
```diff
@@ -257,14 +256,29 @@ pub trait SynthLanguage: Language + Send + Sync + Display + FromOp + 'static {
...
+        time_limit: u64,
+        node_limit: usize,
+    ) -> Ruleset<Self> {
+        synth::synth(SynthParams {
+            workload,
+            prior_rules,
+            iter_limit,
+            time_limit,
+            node_limit,
+        })
```
mwillsey, line 282: You may want some kind of builder pattern or maybe just a `Config` struct of some kind with defaults to allow an easier time configuring ruler runs. It might look something like:
```rust
T::build_run().with_iter_limit(5000).run_workload(workload)
// or 
T::run_workload(workload, Config {
  rules: vec![ /* my prior rules */ ],
  ..Config::default()
)
```

mwillsey (approved):
Looks great! Only one comment about making ruler runs easier to configure, but that can be done later.

PR #89, opened 2022-12-08
Ruleset type

ajpal:


src/enumo/ruleset.rs
```diff
@@ -0,0 +1,38 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.0.is_empty() {
+            None
+        } else {
+            let first = self.0[0].clone();
+            self.0 = self.0[1..].to_vec();
+            Some(first)
+        }
+    }
+}
```
mwillsey, line 24: This is not quite what you are looking for I think, this will cost `n^2` to iterate over because it re-makes the vec everytime.

`Ruleset` is not itself an iterator, but it can be converted into one. This is the same as `Vec`: it's not an iterator itself but it can be converted into one using the [`IntoIterator`](https://doc.rust-lang.org/stable/std/iter/trait.IntoIterator.html) trait. Implement this instead; you can re-use the actual iterator type [`Iter`](https://doc.rust-lang.org/stable/std/slice/struct.Iter.html). Note that `Vec<T>` and `&Vec<T>` have two different implementations of `IntoIterator`: the former yields `T`s and the latter is the same as iterating over a slice, yielding `&T`s.

In this case, the easiest thing is to do neither and just force the user to use the public `.0` field to access the vec. Alternatively, you could impl `IntoIterator for Ruleset<L>` and probably `IntoIterator for &'a Ruleset<L>` as well to mimic `Vec`.

mwillsey (changes requested):

mwillsey (approved):
Looks good now!

PR #90, opened 2022-12-08
ruleset to/from files

ajpal:


src/synth.rs
```diff
@@ -395,19 +393,13 @@ impl<L: SynthLanguage> Synthesizer<L> {
             println!("{:?}      {}", eq.score(), name);
         }
 
-        Report {
```
chandrakananandi, line 398: so what happened to report? we just dont have one?

chandrakananandi (approved):

src/synth.rs
ajpal: Yep, it's gone. We can serialize to/from file but it's just the list of rules (enumo-style), no need for the full json anymore because there aren't really params in the same way.

PR #93, opened 2022-12-14
Trig constant fold

ajpal:


src/language.rs
```diff
@@ -39,15 +40,42 @@ impl<L: SynthLanguage> Analysis<L> for SynthAnalysis {
...
+            rec
+        } else {
+            let mut nodes: Vec<L> = vec![];
+            let mut map: HashMap<Id, Id> = HashMap::default();
+            enode.for_each(|id| {
+                if map.get(&id).is_none() {
+                    let s = get_simplest(&id);
+                    let i = nodes.len();
+                    for n in s.as_ref() {
+                        nodes.push(n.clone().map_children(|id| Id::from(usize::from(id) + i)));
```
chandrakananandi, line 57: not following what's happening here. The `Id::from(usize::from(id) + i)` part specifically.

chandrakananandi (approved):

src/language.rs
chandrakananandi: actually @bksaiki you did this I think originally? what was this for?

bksaiki: Hmmm... good question

bksaiki: I think it's just constructing a RecExpr, but I don't remember why

PR #95, opened 2023-01-02
Don't let time limit be configurable

ajpal:
Time limits are a cause of non-determinism for egg applications. By using a standard 10 minute time limit everywhere, we will reduce the variance caused by timing effects.


chandrakananandi (approved):
Great, thanks so much!

PR #98, opened 2023-01-12
Add egraph to workload enum

ajpal:
Most of this change is the fallout from adding a type parameter to `Workload` so that `EGraph` can be added as a variant.

2 open questions:
1. How do we compare equality of egraphs? 
2. How do we extract the terms represented by an egraph?

Also mostly a question for @mwillsey - is this an acceptable thing to do in Rust? Is there a better way?

### DIFF
--- a/src/enumo/enumosym.rs
+++ b/src/enumo/enumosym.rs
@@ -1,0 +1,44 @@
+use crate::*;
+
+// This is an extremely minimal implementation of SynthLanguage
+// It is not intended for any real uses.
+// Workloads of arbitrary s-expressions can use EnumoSym as a
+// default type parameter.
+egg::define_language! {
+  pub enum EnumoSym {
+    Sym(egg::Symbol),
+  }
+}
+
+impl SynthLanguage for EnumoSym {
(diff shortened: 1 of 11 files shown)


src/enumo/enumosym.rs
```diff
@@ -0,0 +1,44 @@
+use crate::*;
```
mwillsey, line 1: I think you can use egg's [`SymbolLang`](https://docs.rs/egg/latest/egg/struct.SymbolLang.html) instead of this module.

mwillsey (changes requested):
From a Rust perspective, lgtm modulo the note about using `SymbolLang` instead of re-rolling it. But there seem to be some open questions about what an egraph as a workload is/does (see missing eq impl, to egraph impl, presumably some kind of `add_to_egraph` impl as well).

chandrakananandi (approved):
Not much more to add since yesterday's call.

The open questions are quite important and the answers to those may affect whether or not long term we keep this change. Let's discuss them in the next team meeting.

PR #100, opened 2023-01-26
Generalize partition

ajpal:
`partition_sat` was a special case-- this PR makes partition more general.
This allows it to be used for saturating, allowed, and derivability


src/equality.rs
```diff
@@ -103,6 +104,78 @@ impl<L: SynthLanguage> Equality<L> {
...
+            .with_hook(|r| {
+                if r.egraph.find(r.roots[0]) == r.egraph.find(r.roots[1]) {
+                    Err("Done".to_owned())
+                } else {
+                    Ok(())
+                }
+            })
+    }
+
+    pub fn is_derivable(&self, ruleset: &Ruleset<L>, limits: Limits) -> bool {
```
chandrakananandi, line 130: is this file where this method should belong? why?

chandrakananandi (commented):
Looks good! one comment about where `is_derivable` should belong.

src/equality.rs
ajpal: hmm, yeah I thought about this, I could go either way I think. Is it more a property of a rule whether a given ruleset can derive it or a property of a ruleset that it can derive a given rule?

chandrakananandi:
reviewed!

src/equality.rs
chandrakananandi: yeah, that's a great point! I am kinda leaning more towards the latter because I feel like derivability is a quality of a ruleset not a rule? But then, the function should not be called `is_derivable`...

ajpal: Ok I'm fine with that! I'll move it and call it `can_derive`

chandrakananandi: `can_derive` is a good name!

chandrakananandi (approved):
Rest is all great, thanks very much!!

PR #105, opened 2023-02-02
Workload interface

sgpthomas:
Change the workload creation interface so that it doesn't require you to allocate a `Vec`. It instead takes anything that can be turned into an Iterator over `&str`.


ajpal (approved):
Ahh thanks, this is way better! I'm still new to Rust, so I didn't know the right way to do this

PR #107, opened 2023-02-09
Enumo scheduler

ajpal:
Make it easy to swap in different scheduling strategies in enumo recipes

### DIFF
--- a/src/enumo/scheduler.rs
+++ b/src/enumo/scheduler.rs
@@ -1,0 +1,28 @@
+use egg::Runner;
+
+use crate::{Limits, SynthAnalysis, SynthLanguage};
+
+use super::*;
+
+pub enum Scheduler {
+    Simple(Limits),
+}
+
+impl Scheduler {
+    pub fn run<L: SynthLanguage>(
+        &self,
+        runner: Runner<L, SynthAnalysis>,
+        ruleset: &Ruleset<L>,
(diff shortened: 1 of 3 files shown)


src/enumo/scheduler.rs
```diff
@@ -0,0 +1,68 @@
...
+
+                for _ in 0..limits.iter {
+                    // Sat
+                    runner = Self::mk_runner(runner.egraph, &Limits::max()).run(&sat);
+
+                    // Other
+                    runner = Self::mk_runner(
+                        runner.egraph,
+                        &Limits {
+                            iter: 1,
```
chandrakananandi, line 58: oh we only run the other rules for a single iteration? should this be a param? Or will it always be 1?

chandrakananandi: ah I guess it will always be 1

src/enumo/scheduler.rs
```diff
@@ -0,0 +1,68 @@
...
+                            node: limits.node,
```
chandrakananandi, line 59: should we have this limit here? what if the node limit is hit even before iter 1?

chandrakananandi (approved):

src/enumo/scheduler.rs
ajpal: Yeah, I think it makes sense to have- otherwise we never use the node limit

PR #113, opened 2023-03-01
Bidirectional candidates

ajpal:
Most important thing: rules are still just unidirectional! An equality is still just `lhs ==> rhs`
However, there are two changes here:
1. When selecting candidates, if `a ==> b` is chosen and `b ==> a` is also in the candidate set, adds them both at the same time
2. When printing rules, print `a <=> b` instead of `a ==> b` and `b ==> a`

Also a few other changes:
- Or and And filters should take any number of filters, not just two
- rewrite of the rational recipe (gets full equivalence with ruler1)
- add nested fabs to rational recipe (goes deeper than ruler1 could)
- run_workload should be implemented per-domain. this makes it easier to make domain specific changes (like fast_cvec_match vs not) and fits better with the mental model of enumo being a dsl


bksaiki (approved):
Approved!

PR #114, opened 2023-03-03
Parallelize partition

ajpal:


src/enumo/ruleset.rs
```diff
@@ -73,18 +75,20 @@ impl<L: SynthLanguage> Ruleset<L> {
...
+        let results = Mutex::new((Ruleset::default(), Ruleset::default()));
+        let eqs: Vec<&Equality<L>> = self.0.values().collect();
+        eqs.into_par_iter().for_each(|eq| {
+            let f_eq = f(eq);
+            let mut results = results.lock().unwrap();
+            if f_eq {
+                results.0.add(eq.clone());
             } else {
-                outs.add(eq.clone());
+                results.1.add(eq.clone());
```
bksaiki, line 88: Not familiar with Rust multithreading. Is an unlock needed, or will this just do the right thing?

bksaiki (approved):
Left a single comment. Otherwise, it looks good.

chandrakananandi (approved):
So we are processing the rules in parallel for checking if f(r) is true (and partitioning accordingly) right?

maybe check the thing Brett says but looks good!

src/enumo/ruleset.rs
ajpal: I think it just unlocks at the end of the block

PR #116, opened 2023-03-06
Fix `run_workload` for exponential domain

bksaiki:
This PR fixes `run_workload` in the exponential domain. It was erroneously trying to cvec match, but this is a lifting domain.


ajpal (approved):
oops, sorry! this was my b

PR #117, opened 2023-03-06
Move compress and crunch to Scheduler.run

ajpal:


src/enumo/scheduler.rs
```diff
@@ -2,13 +2,14 @@ use std::time::Duration;
...
-use crate::{EGraph, Limits, SynthAnalysis, SynthLanguage};
+use crate::{EGraph, Id, Limits, SynthAnalysis, SynthLanguage};
 
 use super::*;
 
 pub enum Scheduler {
     Simple(Limits),
     Saturating(Limits),
+    Compress(Limits),
 }
```
bksaiki, line 13: Clever. So we have a scheduling concept separate (but related) to egg?

src/enumo/scheduler.rs
```diff
@@ -26,17 +27,19 @@ impl Scheduler {
 
     pub fn run<L: SynthLanguage>(
         &self,
-        egraph: EGraph<L, SynthAnalysis>,
+        egraph: &EGraph<L, SynthAnalysis>,
         ruleset: &Ruleset<L>,
-    ) -> Runner<L, SynthAnalysis> {
+    ) -> EGraph<L, SynthAnalysis> {
```
bksaiki, line 32: Interface nitpick. Why an `EGraph` over a `Runner`? The runner has more information about the run. Maybe we don't need it though.

bksaiki (approved):
Makes sense to me!

src/enumo/scheduler.rs
ajpal: Some runs have multiple phases + multiple runners. Could make sense to return the last runner, I guess, but that's kind of confusing because it seems like it implies that runner would know everything about the run, when it would actually just be for the last phase. I think returning an egraph doesn't have the same issue

PR #120, opened 2023-03-14
Cleanup minimize

ajpal:
since we were always hard-coding step size to 1, let's just rip that out and make everything a little simpler


tests/bool.rs
```diff
@@ -191,7 +191,7 @@ mod test {
         let egraph = Scheduler::Compress(Limits::default()).run(&atoms5.to_egraph(), &all_rules);
         candidates = Ruleset::cvec_match(&egraph);
         let rules5 = candidates.minimize(all_rules.clone(), Limits::default());
-        assert_eq!(rules5.len(), 16);
+        assert_eq!(rules5.len(), 14);
```
oflatt, line 194: Wait so this works better than the old minimizer? Why?

ajpal: not 100% sure, but the 14 rules and 16 rules are equi-derivable, and i'm not sure these tests that assert about the numbers of rules are useful anyways so i didn't think too much about it

oflatt:
Oops, we forgot to merge this and I changed shrink in my branch

ajpal:
> Oops, we forgot to merge this and I changed shrink in my branch

No worries, I think we decided we don't want to use derive instead of shrink anyways because of the performance implications.
I think we do still want the step_size cleanup, but I can just do that separately

PR #123, opened 2023-03-15
Parse bidirectional rules correctly

ajpal:
`a <=> b` should be parsed into two equalities: `a ==> b` and `b ==> a`

### DIFF
--- a/src/equality.rs
+++ b/src/equality.rs
@@ -1,6 +1,6 @@
 use egg::{Analysis, Applier, ENodeOrVar, Language, PatternAst, RecExpr, Rewrite, Subst};
 use std::fmt::Debug;
-use std::{str::FromStr, sync::Arc};
+use std::sync::Arc;
 
 use crate::*;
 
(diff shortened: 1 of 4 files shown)


src/equality.rs
```diff
@@ -12,24 +12,42 @@ pub struct Equality<L: SynthLanguage> {
...
 
-impl<L: SynthLanguage> FromStr for Equality<L> {
-    type Err = String;
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
+impl<L: SynthLanguage> Equality<L> {
+    pub fn from_string(s: &str) -> Result<(Self, Option<Self>), String> {
         if let Some((l, r)) = s.split_once("=>") {
             let l_pat: Pattern<L> = l.parse().unwrap();
             let r_pat: Pattern<L> = r.parse().unwrap();
```
bksaiki, line 19: I think we need to settle on a uniform unidirectional rules format. If we have `a ==> b` and `a => b` in the repo, I'm not sure if this will parse correctly. Separate note, maybe we should try to split on `<=>` first and then do the unidirectional one.

bksaiki (approved):
Left a comment. But otherwise, it makes sense.

PR #127, opened 2023-03-17
New rational recipe

bksaiki:
This PR makes a small change to the rational recipe by tweaking `layer1` to (i) use the constant 2 and (ii) use `run_workload` rather than `run_workload_fast_match`.


bksaiki:
A small change for such a long exploration... Merging should depend on the nightly result.

vee-effekt (approved):

PR #128, opened 2023-03-17
Cleanup

ajpal:
1. rename Equality -> Rule and move into Enumo
2. Remove `apply_unions` which is unused
3. Rename Metric::List to Metric::Lists for consistency


oflatt (approved):
image

PR #129, opened 2023-03-17
add herbie baseline

ajpal:


oflatt:
How is the herbie baseline different than the other rational baseline?

oflatt (approved):
As long as we are sure these are the right ones go for it

ajpal:
> How is the herbie baseline different than the other rational baseline?

These are the rules that Herbie uses (Brett posted them [here](https://uw-cse.slack.com/archives/C01749FQE4U/p1678230473106859)

PR #130, opened 2023-03-20
Use different derivability modes during shrink

oflatt:
This PR allows the shrinking rulesets to use the new derivability modes.
It's very slow though, so only practical for the first iteration.


oflatt:
Merge after merging #124

oflatt:
This PR is on hold for now, merge conflicts and I want to see a nightly to see what it does

src/enumo/ruleset.rs
```diff
@@ -532,56 +538,62 @@ impl<L: SynthLanguage> Ruleset<L> {
...
-        let extract = Extractor::new(&egraph, AstSize);
-        self.0 = Default::default();
-        for (l_id, r_id) in initial {
+        let mut res = Ruleset::default();
+        for (l_id, r_id, eq) in initial {
             if egraph.find(l_id) == egraph.find(r_id) {
                 // candidate has merged (derivable from other rewrites)
                 continue;
             }
-            let (_, left) = extract.find_best(l_id);
```
oflatt, line 550: I think this is maybe a bug? We might not extract out the original rule

src/lib.rs
```diff
@@ -65,7 +66,18 @@ impl Default for Limits {
     fn default() -> Self {
         Self {
             iter: 3,
-            node: 300000,
+            node: 300_000,
+            derive_type: DeriveType::AllRules,
```
oflatt, line 70: Preserves old behavior by default

oflatt:
This PR isn't worth it, I think it does not much

PR #132, opened 2023-03-20
Separate "best recipe" from "replicate ruler1 recipe"

ajpal:
These are two fundamentally different recipes with different goals, so we should make their separation more clear.
 `replicate_ruler1_recipe` should mimic `make_layer` from Ruler1
`best_enumo_recipe` is free to do whatever it likes with guided search, etc


oflatt (changes requested):
I think this means that we now need to test `replicate_ruler1_recipe` and `best_enumo_recipe` against the baseline. Would you please add that benchmark?

PR #133, opened 2023-03-21
Fix filter

ajpal:
I was hitting a bottleneck trying to iterate `[var, const, (uop expr), (bop expr expr), (top expr expr expr)]` up to size 6.
I think this optimization on filters is important: you can actually modify the filter that you push down into the plug. So we used to say something like
`wkld.plug("name", pegs).filter(MetricLt(Atoms, n))`
is the same as
`wkld.plug("name", pegs.filter(MetricLt(Atoms, n)).filter(MetricLt(Atoms, n))`
but you can actually make the filter on the pegs stronger-- in the case of `MetricLt`, it's `n-1` not `n` on the peg

This speeds things up quite a bit.

(It's a little late, so i'm not sure how coherent that explanation is, please lmk if there are questions!)

### DIFF
--- a/src/enumo/workload.rs
+++ b/src/enumo/workload.rs
@@ -99,17 +99,21 @@ impl Workload {
         }
     }
 
-    fn iter(self, atom: &str, n: usize) -> Self {
-        if n == 0 {
-            Self::Set(vec![])
-        } else {
-            let rec = self.clone().iter(atom, n - 1);
-            self.plug(atom, &rec)
+    pub fn pretty_print(&self) {
+        for t in self.force() {
+            println!("{}", t);
         }
     }
 
     pub fn iter_metric(self, atom: &str, met: Metric, n: usize) -> Self {
-        self.iter(atom, n).filter(Filter::MetricLt(met, n + 1))
+        let mut pegs = self.clone();
+        for i in 1..(n + 1) {
+            pegs = self
+                .clone()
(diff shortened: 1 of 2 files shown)


ajpal:
Thanks for the insight @ztatlock! I added a test to demonstrate the issue, and will work on a fix that optimizes the cases we need to make faster without sacrificing soundness in general :)

ajpal:
@oflatt I think this is what you were suggesting to do, but let me know if I misunderstood!

src/enumo/workload.rs
```diff
@@ -247,6 +274,35 @@ mod test {
...
+            "var",
+            "(uop expr)",
+            "(bop expr expr)",
+            "(top expr expr expr)",
+        ]);
+        let three = extended.clone().iter_metric("expr", Metric::Atoms, 3);
+        assert_eq!(three.force().len(), 10);
+
+        let four = extended.clone().iter_metric("expr", Metric::Atoms, 4);
+        assert_eq!(four.force().len(), 32);
```
oflatt, line 289: As long as you checked to make sure it didn't miss anything, this is great!

oflatt:
> @oflatt I think this is what you were suggesting to do, but let me know if I misunderstood!
Not what I was suggesting, but more elegant because it uses plug and filter.

oflatt (approved):

PR #134, opened 2023-03-21
Halide

ajpal:
recipe itself needs work, but I think it would be good to get something merged to main for halide asap- will make it much easier to continue pushing on the derivability results

oopsla-halide.rules comes from running `cargo run --release --bin halide -- synth --iters 1` on https://github.com/uwplse/ruler/commit/acc37eaf4e9a4da2e7a2216aceef45c4d979b695 . 1 iter has 573 nodes and finishes in ~a few seconds. 2 iters has 197778725 nodes and doesn't finish (yay guided search!)


oflatt (approved):
Cool! Is the reduce_monotonic thingy sound?

PR #137, opened 2023-03-21
Fix bad bug in minimize

oflatt:
Minimize re-extracts things for no reason, and this eliminates some rules that should not have been eliminated. This PR fixes that behavior.


ajpal:
Why does re-extracting eliminate rules? I get why it's bad for performance, but I'm not sure I understand why it would lead to different results?

oflatt:
It's because sometimes it extracts stuff with different numbers of variables, then the check fails and it eliminates the rule

ajpal (approved):

PR #138, opened 2023-03-22
Fixed Python script bug.

vee-effekt:
Minor bug in the Python script PR was breaking Nightly (incorrect filepath). This should fix it.


bksaiki (approved):
Looks good!

PR #141, opened 2023-03-23
New tables for baseline, Herbie, Halide results.

vee-effekt:
This PR adds tables for the baseline, Herbie, and Halide results using all the new derivability metrics.


oflatt (commented):
Looks good to me! Though is there a nightly run that is up to date so we can make sure it works?

vee-effekt:
> Looks good to me! Though is there a nightly run that is up to date so we can make sure it works?

This is the most recent one, the only change I've made since then is a minor bugfix to the headers in the html pages with the individual rules, and now raw JSON gets uploaded. I'll kick off a run, though!

http://nightly.cs.washington.edu/reports/ruler/1679589094:nightly:new-tables:836f37ea93/

src/enumo/ruleset.rs
```diff
@@ -281,12 +258,19 @@ impl<L: SynthLanguage> Ruleset<L> {
         }
 
         let stats = json!({
-            "spec": name,
-            "num_rules": num_rules,
-            "num_baseline": baseline.len(),
-            "enumo_derives_oopsla": forwards_derivable,
-            "oopsla_derives_enumo": backwards_derivable,
-            "time": duration.as_secs(),
+            "domain": name,
```
ajpal, line 261: Instead of making the json fields baseline-specific, can they always be the same? That will simplify the to-html and to-latex scripts

src/enumo/ruleset.rs
```diff
@@ -314,35 +298,53 @@ impl<L: SynthLanguage> Ruleset<L> {
...
+
+        let start_f = Instant::now();
+        let (can_f, cannot_f) = self.derive(derive_type, &baseline, limits);
+        let time_f = start_f.elapsed();
+        let start_b = Instant::now();
+        let (can_b, cannot_b) = baseline.derive(derive_type, self, limits);
+        let time_b = start_b.elapsed();
+
+        let derivability_results = json!({
+            "forwards derivable": &can_f.to_str_vec(),
```
ajpal, line 332: Don't use "forwards" and "backwards" anywhere. As we've talked about, that wording can be confusing so it's better to say `enumo_derives_baseline` or `baseline_erives_enumo`

src/enumo/ruleset.rs
ajpal: keys `[baseline_name, enumo_spec_name, num_rules, num_baseline, enumo_derives_baseline, baseline_derives_enumo, time, domain]` would be good I think

PR #142, opened 2023-03-23
Add script to parse json data to latex table

ajpal:
data should be in the format
```
[
  "baseline_name": [
    {
        "domain": X,
        "# rules found": X,
        "rulefinding time (sec)": X,
        "# <baseline_name> rules": X,
        "enumo -> <baseline_name> (lhs, lhs & rhs, all)": X, X, X,
        "enumo -> <baseline_name> time": X, X, X,
        "<baseline_name> -> enumo (lhs, lhs & rhs, all)": X, X, X,
        "<baseline_name> -> enumo time": X, X, X,
        "minimization strategy": X
    },
    ...
  ],
  ...
]
```
And output looks like:
image


scripts/generate_latex_tables.py
```diff
@@ -0,0 +1,93 @@
+#!/usr/bin/python3
+import sys
+import json
+
+def get_headers_for_baseline(domain):
+  return [
+    ("domain", "Domain"),
+    ("# rules found", "SlideRule Rules"),
+    ("rulefinding time (sec)", "Time (s)"),
+    ("# {} rules".format(domain), "Baseline Rules"),
```
oflatt, line 10: This is a bit hacky but I suppose it works, just need to make sure we change this and the json in sync

oflatt (approved):
Great!

PR #144, opened 2023-03-27
Add is_valid as public function on Rule

ajpal:
This makes writing the formalism easier

(Also renames `eq` to `rule` everywhere, this is just a little nit cleanup from when I renamed `Equality` to `Rule`)


chandrakananandi (approved):
Great!! Also need to add this to Section 4.

PR #146, opened 2023-03-27
Make minimize take Scheduler

ajpal:
This makes writing formalism easier/better 

(Also probably worth doing experiments with different scheduling techniques in minimize)


chandrakananandi (approved):
Perfect. I realize that I made a mistake in the formalism. The operators like `minimize`, `can_derive`, and `eq_sat` should *not* also take a `limit`. If you are looking at that section, feel free to correct it. I can do it later too.

PR #147, opened 2023-03-27
Fix nightly, again

bksaiki:
This PR restructures the nightly infrastructure in an attempt to fix persistent issues with nightlies. All nightlies scripts and resource files will be stored under `infra/` while reports are generated under `nightly/`. In particular,

`infra/nightly.sh` - top-level script to invoke nightlies; called by the Makefile
`infra/generatehtml.py` - reads json files in a partially-generated nightly report under `nightly/` to create an html report
`infra/nightly/` - directory containing templates and other resource files for nightly reports

`nightly/json/` - output directory for Rust tests generated by `cargo test ...` and the like
`nightly/output/` - output location of the nightly artifact

Hopefully, this finally solves issues.


infra/nightly.sh
```diff
@@ -0,0 +1,55 @@
...
+  dir="$(cd -P "$(dirname "$src")" && pwd)"
+  src="$(readlink "$src")"
+  [[ $src != /* ]] && src="$dir/$src"
+done
+MYDIR="$(cd -P "$(dirname "$src")" && pwd)"
+
+# Absolute directory paths
+TOP_DIR="$MYDIR/.."
+RESOURCE_DIR="$MYDIR/nightly"
+NIGHTLY_DIR="$TOP_DIR/nightly"
```
vee-effekt, line 23: Could possibly be a good idea to call these different things, perhaps nightly_resources and nightly_output or something.

bksaiki: Yep, probably a good idea to at least rename one of them.

vee-effekt (changes requested):
Looks good aside from the minor change I proposed.

chandrakananandi (approved):
:shipit:

PR #150, opened 2023-03-30
Change rational rule semantics to preserve error, resulting in more rules

oflatt:
Previously, z3 was free to choose a particular value for division by zero. But since it has to choose just one value, this filtered out good rules.
Now, we form the query to check 1) equality assuming no division by zero and 2) that division by zero is preserved.

This PR also adds a method to learn conditional rules, but doesn't use it yet.


src/enumo/ruleset.rs
```diff
@@ -79,6 +79,12 @@ impl<L: SynthLanguage> Ruleset<L> {
         Ruleset(map)
     }
 
+    pub fn union(&self, other: &Self) -> Self {
```
ajpal, line 82: Do we need this? `Ruleset.extend` already exists

tests/rational.rs
```diff
@@ -126,12 +136,29 @@ impl SynthLanguage for Math {
...
-            z3::SatResult::Unsat => ValidationResult::Valid,
-            z3::SatResult::Unknown => ValidationResult::Unknown,
-            z3::SatResult::Sat => ValidationResult::Invalid,
+        let lhs_denom = Self::all_denominators(Self::pat_to_sexp(lhs));
+        let rhs_denom = Self::all_denominators(Self::pat_to_sexp(rhs));
+        let denominators = lhs_denom.union(&rhs_denom);
+
+        let mut assert_equal = lexpr._eq(&rexpr);
+        let zero_pat = "0".parse::<Pattern<Math>>().unwrap();
+        let zero_z3 = egg_to_z3(&ctx, Self::instantiate(&zero_pat).as_ref());
```
ajpal, line 145: I think `let zero = z3::ast::Real::from_real(ctx, 0, 1);` would be an easier way to define zero

ajpal (approved):

PR #151, opened 2023-03-30
Refactor Nightly Scripts

ajpal:
- Pull writing to json logic out of Ruleset and into its own module
- Parse data into html tables and latex tables in JS on the frontend
- Reduce duplication for each domain

New UI can be seen here: http://nightly.cs.washington.edu/reports/ruler/1680195670:nightly:logger:fd4ff07750/


vee-effekt (approved):
Looks fantastic! The Halide table / LaTeX isn't loading properly, but I'm guessing the last commit you pushed fixes that?

PR #156, opened 2023-04-03
Consolidate Append

ajpal:
Little optimization to combine appends when possible

### DIFF
--- a/src/enumo/workload.rs
+++ b/src/enumo/workload.rs
@@ -157,7 +157,14 @@ impl Workload {
     }
 
     pub fn append(self, workload: impl Into<Workload>) -> Self {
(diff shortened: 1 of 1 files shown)


src/enumo/workload.rs
```diff
@@ -157,7 +157,14 @@ impl Workload {
     }
 
     pub fn append(self, workload: impl Into<Workload>) -> Self {
-        Workload::Append(vec![self, workload.into()])
+        match self {
```
oflatt, line 160: What if workload is an append? Wouldn't you want to consolidate with that one as well?

oflatt (approved):
I think you should pattern match on both independently and add them to some vector
But I approve

PR #159, opened 2023-04-04
Make conditional workload more enumo-y

ajpal:


oflatt:
Great change, but I think this means we need to fix the z3 encoding to work with if statements and division... I can work on that if you want, or you can give it a shot.

I think our idea was fall back on the old validity checker if you detect an if statement

oflatt:
I ended up implementing a better verifier in my branch that handles ifs... PR coming soon

ajpal:
covered by https://github.com/uwplse/ruler/pull/175

PR #160, opened 2023-04-04
Add div to recipe

ajpal:
Adding more stuff to the cvec seems to solve the cvec crashes we were seeing earlier.
my guess is that since `0` and `1` are also used as booleans in this domain, we need to have more variety in the cvec to prevent merging constants


oflatt:
I'm confused, are we merging based on cvecs? Is that where the unsoundness came from?
My mental model was that the cvecs are a heuristic for finding rules, which we verify

oflatt:
Should we merge but open a issue with a repro of the problem?

ajpal:
https://github.com/uwplse/ruler/issues/167

PR #162, opened 2023-04-05
preserve var order

ajpal:
A couple unrelated bug fixes:
1. just a simple arithmetic bug in the interval analysis for nat
2. we should initialize variables in the order they appear in the workload. I'm still not totally clear on why this makes a difference, but since it seems to, we should eliminate this source of non-determinism
3. Add reverse direction rules properly. Previously, I was calling `from_recexprs(e1, e2)`  and `from_recexprs(e2, e1)`, but `from_recexprs` generalizes the RecExpr to a pattern, so it was not actually giving the right results. Like, `from_recexpr("(+ b a)", "(+ a b)")` returns the rule `(+ ?b ?a) ==> (+ ?a ?b)` but `from_recexpr("(+ a b)", "(+ b a)")` _also_ returns the rule `(+ ?b ?a) ==> (+ ?a ?b)` since it re-generalizes the recexprs from scratch. What we actually want is to just generalize the recexprs once, and then make both directions of rules using those patterns.

### DIFF
--- a/tests/nat.rs
+++ b/tests/nat.rs
@@ -93,7 +93,7 @@ impl SynthLanguage for Nat {
                     .high
                     .clone()
                     .zip(y_int.high.clone())
-                    .map(|(a, b)| a + b);
+                    .map(|(a, b)| a * b);
 
                 Interval::new(Some(low), high)
             }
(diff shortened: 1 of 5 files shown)


tests/nat.rs
```diff
@@ -93,7 +93,7 @@ impl SynthLanguage for Nat {
                     .high
                     .clone()
                     .zip(y_int.high.clone())
-                    .map(|(a, b)| a + b);
+                    .map(|(a, b)| a * b);
```
oflatt, line 96: 😱

oflatt (approved):
Good changes

ajpal:
subsumed by https://github.com/uwplse/ruler/pull/168

PR #166, opened 2023-04-05
re-add short circuiting

ajpal:
We should short-circuit running eqsat when we're checking derivability- as soon as the lhs and rhs merge, we're done.
We had this previously, it was removed in https://github.com/uwplse/ruler/pull/107, but we should re-add it. I think it will make a big difference in derivability performance


chandrakananandi (approved):
:shipit:

PR #168, opened 2023-04-06
Cleanup recipes

ajpal:


src/lib.rs
```diff
@@ -78,6 +79,22 @@ impl Limits {
...
+
+    // match oopsla limits
+    pub fn rulefinding() -> Self {
+        Self {
+            iter: 2,
+            node: 300_000,
+        }
+    }
+
+    // match oopsla limits
```
vee-effekt, line 91: These aren't the OOPSLA limits, though, right? Are we using 3 iters for eval?

tests/rational.rs
```diff
@@ -766,76 +735,42 @@ pub mod test {
...
-            &herbie,
-            "rational_best",
-            "herbie",
-            Limits {
-                iter: 2,
-                node: 150000,
-            },
-            duration,
-        );
+        logger::write_output(&best_rules, &ruler1, "rational_best", "oopsla", duration);
```
vee-effekt, line 752: Are we no longer interested in the replicate rules?

tests/recipes/rational_replicate.rs
```diff
@@ -1,36 +1,51 @@
...
     let mut rules = Ruleset::default();
     let limits = Limits::default();
 
     // Domain
-    let lang = Workload::new(&["var", "const", "(uop expr)", "(bop expr expr)"]);
     let vars = &Workload::new(["a", "b", "c"]);
     let consts = &Workload::new(["0", "-1", "1"]);
     let uops = &Workload::new(["~", "fabs"]);
     let bops = &Workload::new(["+", "-", "*", "/"]);
+    let lang = base_lang()
```
vee-effekt, line 16: Any reason we're using Atoms-based enumeration on the other 3 recipes, but still doing the layering thing here? Doesn't matter to me as long as the results match up, just wondering.

src/lib.rs
```diff
@@ -78,6 +79,22 @@ impl Limits {
...
+    pub fn deriving() -> Self {
```
vee-effekt, line 92: Are we planning to use 3 iters (as opposed to 5) for eval now?

vee-effekt (approved):
Approved—few questions in the comments, but looks good to me.

tests/rational.rs
ajpal: no we still want all four- they're all here?

tests/recipes/rational_replicate.rs
ajpal: no real reason- I vaguely remember atoms not as well last fall, but i didn't try it tbh

src/enumo/scheduler.rs
```diff
@@ -99,4 +133,33 @@ impl Scheduler {
...
+    pub fn run<L: SynthLanguage>(
+        &self,
+        egraph: &EGraph<L, SynthAnalysis>,
+        ruleset: &Ruleset<L>,
+    ) -> EGraph<L, SynthAnalysis> {
+        // Set timeout to something quite high (10 minutes) to reduce non-deterministic
+        // behavior. If egg terminates due to timeout, it's hard to reason about
+        // what did/didn't merge, and things vary a lot on different machines, so
+        // it is best to avoid terminating due to timeout.
+        self.run_internal(egraph, ruleset, None, Duration::from_secs(600))
```
vee-effekt, line 146: May need to up this for Halide / further parameterize it at some point, but that's a small change so let's just do this for now.

vee-effekt (approved):
Looks good!

PR #170, opened 2023-04-06
Skip nightlies recipes on CI

ajpal:
These tests have no assertions anyways, so there's not really a point to running them in the CI. They really only matter for nightlies

~DO NOT MERGE until there's a nightlies run of this branch to make sure that all of these tests DO run on nightly.~
Nightlies run looks fine!


vee-effekt (approved):
Looks good.

PR #171, opened 2023-04-06
Make sure cvecs are always initialized properly

ajpal:
I got impatient so it hasn't finished running yet, but I think this will fix the soundness issues you've been having with conditionals. Please holler if it doesn't!

### DIFF
--- a/tests/bool.rs
+++ b/tests/bool.rs
@@ -85,8 +85,6 @@ impl SynthLanguage for Bool {
     }
 
     fn initialize_vars(egraph: &mut EGraph<Self, SynthAnalysis>, vars: &[String]) {
-        println!("initializing vars: {:?}", vars);
-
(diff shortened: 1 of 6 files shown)


tests/bool.rs
```diff
@@ -358,7 +356,7 @@ mod test {
                 node: 1000000,
             },
         );
-        assert_eq!(can.len(), 12);
-        assert_eq!(cannot.len(), 12);
+        assert!(can.len() > 0);
```
oflatt, line 359: :eye:
why

oflatt: nvm

oflatt (approved):
yay Anjali saves the day

oflatt:
Do we still need this?

ajpal:
https://github.com/uwplse/ruler/pull/172 this is a better approach

PR #178, opened 2023-04-08
Fix halide baseline rules

ajpal:
lmao we just weren't parsing these right


chandrakananandi (approved):
LOL great catch 😂

vee-effekt:
What does this mean exactly? Are the current results wrong?

ajpal:
> What does this mean exactly? Are the current results wrong?

This won't change the results using enumo rules to derive the halide ones.
But without the `?` in the halide rules, the variables weren't being parsed as pattern variables (so for example the rule `(! (! x)) ==> x` would only match a variable enode literally named `x`). The halide->enumo derivability was saying 0 for everything (and I think we always knew this was suspect but just hadn't had a chance to dig into). This should fix that direction.
Screenshot 2023-04-08 at 10 13 24 PM

vee-effekt:
ooooooooo okay great

PR #179, opened 2023-04-09
Take Halide off.

vee-effekt:
Halide takes too long on the nightlies, and the most up-to-date recipe won't finish using Nightly's specs anyway, so let's resign ourselves to running it locally.


ajpal:
I don't love that we have to do this because it'll make our halide results a little harder to keep track of, but I guess we don't really have a choice.
Is it just the derivability that takes too long? if so, it'd be great to keep the recipe/rulefinding on nightly so at least we know our enumo stuff is all working. (fine to merge this now and do as follow up though!!)

vee-effekt:
Unfortunately, both take a pretty long time, but derivability takes much longer. I can time rulefinding tonight.

ajpal:
merging now before nightlies starts again :)

ajpal (approved):

PR #181, opened 2023-04-10
Use i64 instead of Bigint

ajpal:


src/enumo/workload.rs
```diff
@@ -81,9 +82,15 @@ impl Workload {
         }
         L::initialize_vars(&mut egraph, &vars);
 
+        let start = Instant::now();
         for sexp in sexps.iter() {
             egraph.add_expr(&sexp.to_string().parse::<RecExpr<L>>().unwrap());
         }
+        println!(
```
vee-effekt, line 89: maybe remove this but also whatever

vee-effekt (approved):
One comment, but looks good.

ajpal:
subsumed by https://github.com/uwplse/ruler/pull/182

PR #182, opened 2023-04-10
Speed up derivability

ajpal:


src/logger.rs
```diff
@@ -134,6 +127,7 @@ pub fn get_derivability_results<L: SynthLanguage>(
     let time_f = start_f.elapsed();
     let start_b = Instant::now();
     let (can_b, cannot_b) = baseline.derive(derive_type, ruleset, limits);
+    println!("{} / {} ({})", can_b.len(), cannot_b.len(), ruleset.len());
```
vee-effekt, line 130: Maybe remove this.

vee-effekt (approved):
Looks good.

mwillsey:
1000 feels extremely low, search_limit is intended as a last ditch effort to stop e-matching, I don't think you wanna be hitting it in "normal" use cases. But if it works it works :shrug:

PR #183, opened 2023-04-11
Clean up how we add candidates from recexprs

ajpal:
this fixes the bug @oflatt found and also consolidates this logic in one place so we won't have to change it in multiple places again if there's a bug.


oflatt (approved):
Yayyyy much better

PR #187, opened 2023-04-12
Add Szalinski eval; parameterize match limits

rtjoa:


src/enumo/ruleset.rs
```diff
@@ -491,7 +491,7 @@ impl<L: SynthLanguage> Ruleset<L> {
             }
         }
 
-        let out_egraph = scheduler.run_derive(&egraph, self, rule);
+        let out_egraph = scheduler.run_internal(&egraph, self, Some(rule));
```
ajpal, line 494: wait, let's not get rid of `run_derive`, we still want that

rtjoa: `run_derive` use to just call `run_internal` with a match limit of 1000. Now, the match limit is in the scheduler.

I'll leave `run_derive` in, and have it just wrap `rule` in `Some`?

ajpal: yep

mwillsey (approved):
Looks good to me! (note: I only reviewed the Sz stuff)

chandrakananandi (approved):
Nice work!

rtjoa:
@ajpal Am I good to merge?

PR #175, opened 2023-04-08
Learn conditional factoring rules in rational

oflatt:
This PR significantly improves the rational suite by adding a recipe that learns factoring rules.
It also changes the constant folder from `Ratio<BigInteger>` to `Rational64` for performance.

PR #188, opened 2023-04-12
Halide recipe that gets 64% derivability.

vee-effekt:
Also, changes to the logger so you can choose not to attempt derivability.

### DIFF
--- a/src/enumo/scheduler.rs
+++ b/src/enumo/scheduler.rs
@@ -137,7 +137,7 @@ impl Scheduler {
         egraph: &EGraph<L, SynthAnalysis>,
         ruleset: &Ruleset<L>,
     ) -> EGraph<L, SynthAnalysis> {
-        self.run_internal(egraph, ruleset, None, 200_000)
+        self.run_internal(egraph, ruleset, None, 1_000_000)
     }
 
     pub fn run_derive<L: SynthLanguage>(
--- a/src/enumo/scheduler.rs
+++ b/src/enumo/scheduler.rs
@@ -137,7 +137,7 @@ impl Scheduler {
         egraph: &EGraph<L, SynthAnalysis>,
         ruleset: &Ruleset<L>,
     ) -> EGraph<L, SynthAnalysis> {
-        self.run_internal(egraph, ruleset, None, 1_000_000)
+        self.run_internal(egraph, ruleset, None, 200_000)
     }
 
     pub fn run_derive<L: SynthLanguage>(
--- a/src/logger.rs
+++ b/src/logger.rs
@@ -17,18 +17,23 @@ pub fn write_output<L: SynthLanguage>(
 ) {
     // get information about the derivability of our ruleset vs. the baseline ruleset
     let start = Instant::now();
+    let halide = if recipe_name == "halide" {
+        true
+    } else {
+        false
+    };
     let ((forwards_lhs, backwards_lhs), (lhs_f, lhs_b), results_lhs) =
-        get_derivability_results(ruleset, DeriveType::Lhs, baseline);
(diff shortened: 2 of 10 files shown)


src/enumo/scheduler.rs
```diff
@@ -137,7 +137,7 @@ impl Scheduler {
         egraph: &EGraph<L, SynthAnalysis>,
         ruleset: &Ruleset<L>,
     ) -> EGraph<L, SynthAnalysis> {
-        self.run_internal(egraph, ruleset, None, 200_000)
```
ajpal, line 140: why did you change this back? @oflatt  lowered it in https://github.com/uwplse/ruler/pull/175

src/logger.rs
```diff
@@ -112,33 +113,47 @@ pub fn get_derivability_results<L: SynthLanguage>(
...
-    });
-
-    (
-        (can_f.len(), can_b.len()),
-        (time_f, time_b),
-        derivability_results,
-    )
+        let limits = if let DeriveType::AllRules = derive_type {
+            Limits {
+                iter: 2,
```
ajpal, line 132: since we're not doing derive for halide on nightly anymore, can we put this back at 5?
or are we getting rid of All entirely?

tests/recipes/halide.rs
```diff
@@ -70,6 +61,53 @@ pub fn halide_rules() -> Ruleset<Pred> {
         ),
         all_rules.clone(),
     );
-    all_rules.extend(full);
+    println!("full finished.");
+    all_rules.to_file("thru_full.rules");
```
ajpal, line 65: take out these write to files

tests/halide.rs
```diff
@@ -322,12 +319,15 @@ mod test {
...
 
+        logger::write_output(&all_rules, &baseline, "halide", "halide", duration, false);
+
         logger::write_output(
             &oopsla_halide,
             &baseline,
             "oopsla halide (1 iter)",
             "halide",
             oopsla_duration,
+            true,
```
ajpal, line 330: make this false too

src/logger.rs
vee-effekt: I think we're getting rid of it entirely.

src/enumo/scheduler.rs
vee-effekt: I'll put it back, wanted to see what would happen.

tests/halide.rs
vee-effekt: oh, why?

src/logger.rs
```diff
@@ -112,33 +113,47 @@ pub fn get_derivability_results<L: SynthLanguage>(
...
     derive_type: DeriveType,
     baseline: &Ruleset<L>,
+    derive: bool,
 ) -> ((usize, usize), (Duration, Duration), Value) {
-    let limits = if let DeriveType::AllRules = derive_type {
-        Limits {
-            iter: 2,
-            node: 100_000,
-        }
+    if !derive {
```
ajpal, line 118: I think it's weird/confusing to have this check in this function. it would make more sense to have it in `write_output`. Like, we shouldn't be calling a function called `get_derivability_results` if we don't want to actually run derivability

tests/halide.rs
ajpal: if we're not running derivability for the enumo halide recipe, there's no point running it for the baseline. i think it'll make it more clear in the nightly report that it just didn't run if everything is 0s

src/logger.rs
vee-effekt: it's a lot easier to write this way

vee-effekt: but i'll change it

PR #190, opened 2023-04-12
Halide tweaks.

vee-effekt:
Okay, I actually don't know how to revert a commit on Git, so this is what I came up with—sorry!

### DIFF
--- a/tests/recipes/halide.rs
+++ b/tests/recipes/halide.rs
@@ -1,17 +1,13 @@
 use ruler::{
-    enumo::{Metric, Ruleset, Filter, Workload},
-    recipe_utils::{recursive_rules, Lang, run_workload},
-    Limits
+    enumo::{Filter, Metric, Ruleset, Workload},
+    recipe_utils::{recursive_rules, run_workload, Lang},
+    Limits,
 };
 
 use crate::Pred;
 
 pub fn halide_rules() -> Ruleset<Pred> {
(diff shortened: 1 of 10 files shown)


tests/recipes/halide.rs
```diff
@@ -24,23 +21,19 @@ pub fn halide_rules() -> Ruleset<Pred> {
...
-    // Rational rules up to size 5:
     let rat_only = recursive_rules(
         Metric::Atoms,
         5,
         Lang::new(
             &["-1", "0", "1"],
             &["a", "b", "c"],
             &["-"],
-            &["+", "-", "*", "/", "min", "max"],
+            &["+", "-", "*", "min", "max"],
```
ajpal, line 31: no more div? why? if it just blows things up too much then maybe we should have it here but not in the `all` workload? it's hard to imagine that learning simple arithmetic rules with div could be *worse*?

vee-effekt: it's not worse, but it gets you nothing and adds like 300 rules

vee-effekt: difference between 910 and 1232 rules. I feel like 910 looks better if proving power is the same

ajpal: whoa that's a big difference. ok leave it out.

ajpal (approved):

PR #196, opened 2023-04-12
Update UI + latex tables

ajpal:
Display derivability as percentage
image

Display missing rules bidirectionally on bv experiment
image

Generate bv latex
image

reformatting on tables
image


vee-effekt (approved):
Looks great. These will look much better in the paper, I think.

PR #197, opened 2023-04-13
This is the 75% Halide recipe.

vee-effekt:
Don't do anything with it. Just for posterity.


tests/halide.rs
```diff
@@ -330,4 +335,132 @@ mod test {
             false,
         )
     }
+
+    // Uncomment to run the fancy 75% recipe.
```
ajpal, line 339: Can you add to the  comment explaining that this recipe can only run on leviathan and that's why it's commented out

vee-effekt: Done.

PR #199, opened 2023-04-13
BV4 baseline reversion.

vee-effekt:
The new one that's fancy is still there but now we have the old, dumb one, too.

### DIFF
--- a/tests/bv128.rs
+++ b/tests/bv128.rs
@@ -4,19 +4,19 @@
 
 ruler::impl_bv!(128);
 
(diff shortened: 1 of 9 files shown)


ajpal:
The other bv files import the bv4 rules (ex: https://github.com/uwplse/ruler/blob/main/tests/bv64.rs#L67)
I think with your change they will now import the simpler recipe, which is not what we want. Maybe make a new file under `tests/recipes` (similar to how we have two versions for rational) so that we can import the right thing from the other bv tests

ajpal (approved):

PR #202, opened 2023-04-13
lhs and lhsrhs for bv

ajpal:
run both lhs and lhs-rhs derivability
also include time
image

### DIFF
--- a/infra/nightly-resources/generate.js
+++ b/infra/nightly-resources/generate.js
@@ -233,36 +233,38 @@ function onGenerateClick(version) {
 }
 
 function generateLatex(version, tableData, ignoreColumns) {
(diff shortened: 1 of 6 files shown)


vee-effekt (approved):
Are we just taking out the Latex because we're finished with baseline / Herbie eval tables?

ajpal:
> Are we just taking out the Latex because we're finished with baseline / Herbie eval tables?

see https://uw-cse.slack.com/archives/C01749FQE4U/p1681418179967689. we've manually changed the formatting (column names, ordering, etc) so i don't want people to drop in the generated latex and break things
I will go back and update the auto-gen but not before deadline

PR #208, opened 2023-04-25
Update rational_best for the herbie eval

oflatt:
This PR simplifies rational_best, which makes herbie perform better
The branch is called "unsound" because the z3 verifier still allows dropping variables (which is unsound)


src/enumo/scheduler.rs
```diff
@@ -137,7 +137,7 @@ impl Scheduler {
         egraph: &EGraph<L, SynthAnalysis>,
         ruleset: &Ruleset<L>,
     ) -> EGraph<L, SynthAnalysis> {
-        self.run_internal(egraph, ruleset, None, 200_000)
+        self.run_internal(egraph, ruleset, None, 1_000_000)
```
oflatt, line 140: reverted this sus change

tests/rational.rs
```diff
@@ -740,26 +817,49 @@ pub mod test {
         let start = Instant::now();
         let best_rules = best_enumo_recipe();
         let duration = start.elapsed();
+        // herbie and ruler can't learn any if rules
+        // but crash if you include them
+        let without_if: Ruleset<Math> = Ruleset::new(
```
oflatt, line 816: Importantly, we have to leave out the div rules

src/lib.rs
```diff
@@ -88,11 +88,10 @@ impl Limits {
         }
     }
 
-    // match oopsla limits
     pub fn deriving() -> Self {
         Self {
             iter: 5,
-            node: 100_000,
```
ajpal, line 95: I think it's important that we run with the same limits as Oopsla in order for it to be a fair comparison, I don't think we should change this.

tests/rational.rs
```diff
@@ -793,10 +887,9 @@ pub mod test {
 
         let guarded_rules = run_workload(terms, all_rules.clone(), Limits::rulefinding(), false);
         guarded_rules.to_file("guard.rules");
-        assert!(guarded_rules.0.contains_key("(if ?a 1 (/ 1 ?a)) ==> 1"));
```
ajpal, line 796: Wait, does this mean we don't learn this rule anymore? That's not good.

oflatt: I made the recipe simpler so it doesn't find if rules like these

tests/rational.rs
```diff
@@ -283,6 +312,53 @@ impl Math {
...
+                    let new_path_neg = z3::ast::Bool::and(&ctx, &[&path, &cond_real._eq(&zero)]);
+                    res.extend(Self::error_conditions(ctx, list[2].clone(), new_path_pos));
+                    res.extend(Self::error_conditions(ctx, list[3].clone(), new_path_neg));
+                } else {
+                    for s in list {
+                        res.extend(Self::error_conditions(ctx, s, path.clone()));
+                    }
+                };
+            }
+            Sexp::String(_atom) => (),
```
ajpal, line 355: nit: _ if unused

tests/rational.rs
```diff
@@ -283,6 +312,53 @@ impl Math {
         }
     }
 
+    fn error_conditions<'a>(
```
ajpal, line 315: can you write a doc comment for this function since it's a little complex?

ajpal (approved):
lgtm, a few little comments to resolve before merging!
also make sure to merge in main because I merged my cleanup PR so you might need to resolve conflicts

oflatt:
Thanks for the review, fixed it up @ajpal

oflatt:
Spoke too soon, need to fix up a test

PR #209, opened 2023-04-27
Restructure Nightlies JSON

ajpal:
- Major refactor of the JS code to make it better/cleaner
- Slight changes to the backend logger code
- Change to the data format to reduce redundancy and add some more structure to the JSON objects


infra/nightly-resources/latex.js
```diff
@@ -0,0 +1,195 @@
...
+  lines = lines.concat(getCaption(tableName));
+  lines.push(String.raw`\label{table:${tableName}}`);
+  lines.push(String.raw`\end{table}`);
+
+  let elem = document.getElementById("latex");
+  elem.innerHTML = lines.join("\n");
+  elem.style.height = "200px";
+}
+
+function getCaption(version) {
```
oflatt, line 149: We should pass in the string from the outside instead of casing on the version here

oflatt (approved):
Great refactor, much needed

infra/nightly-resources/latex.js
ajpal: Sorry, what do you mean?

oflatt: Oh what I meant is
Do we do this check on the version anywhere else? It would be good to bundle all the specific stuff together in the code base

PR #225, opened 2023-10-13
Rename everything to fast-forwarding

ajpal:
Now that the paper is done and everything is settled, we should remove references to "lifting", "allow/forbid", etc and call everything fast-forwarding


oflatt:
Sorry forgot to review but good change

PR #231, opened 2025-04-21
Graceful handling for invalid workload terms

ajpal:
Previously we didn't need to worry as much about invalid terms (basically, we put the onus on workload writers to enumerate terms correctly)
However, I'm going to be doing some experiments with LLM generated workloads, and the LLMs introduce all kinds of incorrect terms, making it more important to gracefully handle malformed input.


tests/bool.rs
```diff
@@ -294,6 +294,14 @@ mod test {
         logger::write_baseline(&rules, "bool", &baseline, "oopsla", duration);
     }
 
+    #[test]
+    fn parse_invalid_wkld() {
+        let wkld = Workload::from_file("llm/out/bool.wkld");
```
chandrakananandi, line 299: should this file be added as part of this PR? How will this test run otherwise?

src/enumo/workload.rs
```diff
@@ -48,12 +48,26 @@ impl Workload {
         let infile = std::fs::File::open(filename).expect("can't open file");
         let reader = std::io::BufReader::new(infile);
         let mut sexps = vec![];
-        for line in std::io::BufRead::lines(reader) {
-            sexps.push(line.unwrap().parse().unwrap());
+        for line in std::io::BufRead::lines(reader).flatten() {
+            if let Ok(sexp) = line.parse() {
+                sexps.push(sexp);
+            } else {
+                println!("Skipping invalid s-expression: {}", line);
```
chandrakananandi, line 55: So we just want to move on to the other expressions and skip invalid ones. We intentionally don't want to throw an exception here, right?

PR #234, opened 2025-05-20
Fixing Rust warning, build error, and outdated aliases

bksaiki:
Addresses a few problems:
- #232 updated the Rust toolchain; the build complains about `.cargo/config` which is now deprecated => changed to `.cargo/config.toml`
- cannot build Z3 with a modern Python version => updated from `0.10.0` to `0.12.1`
- aliases pointed to `bin/` => now runs inline tests under `tests/`


ajpal (approved):
thanks! lgtm once CI passes

PR #240, opened 2026-05-26
Upgrade z3 0.12.1 → 0.20

ajpal:
(Code by Claude)

API migration:
- Cargo feature `static-link-z3` was renamed `bundled`.
- AST types lost their lifetime parameter: `z3::ast::Int<'a>` → `z3::ast::Int`
- All AST constructors no longer take `&Context`: `Int::from_i64(&ctx, 0)` → `Int::from_i64(0)` `BV::new_const(&ctx, name, n)` → `BV::new_const(name, n)` `Real::from_real(&ctx, 1, 2)` → `Real::from_real(1, 2)` `Bool::from_bool(&ctx, true)` → `Bool::from_bool(true)`
- Variadic ops drop the `&Context` argument: `Int::add(&ctx, &[&a, &b])` → `Int::add(&[a.clone(), b.clone()])` Same for `Bool::and` / `Bool::or` / `Real::add` / etc. The slice element bound changed to `T: Into<Self> + Clone`, so we clone where we previously passed `&T`.
- Binops are now methods on the receiver instead of free functions: `Int::lt(l, r)` → `l.lt(r)` `Bool::ite(&cond, &a, &b)` → `cond.ite(&a, &b)`
- `Solver::new(&ctx)` → `Solver::new()` (no args).
- `Context::new(&cfg)` is gone; wrap solver work in `z3::with_z3_config(&cfg, || { ... })`, which scopes a thread-local context for the closure body.
- `_eq` is deprecated in favor of `eq` (still works with a warning).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,8 +21,7 @@ float-cmp = "0.8.0"
 lazy_static = "1.4.0"
 itertools = "0.9.0"
 # fixed = "1.9.0"
-#z3 = "0.12.1"
-z3 = {version = "0.12.1", features = ["static-link-z3"]}
+z3 = {version = "0.20", features = ["bundled"]}
 rustc-hash = "1"
 symbolic_expressions = "5"
 rayon = "1"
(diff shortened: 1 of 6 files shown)


chandrakananandi (approved):
LGTM!

